PR #9, opened 2022-10-11
Add guidelines for guidelines

mcmire:
I feel that it would be good to have some amount of standardization around the documents that we add to this repo, at least when it comes to guidelines documents. These kinds of documents are basically souped up "do"s and "don't"s, and so it seems that we can assign a structure to them.

If we want something like this and feel that this belongs in `guides/README.md` then I'm happy to move it.

### DIFF
--- a/guides/guidelines.md
+++ b/guides/guidelines.md
@@ -1,0 +1,69 @@
+# Guidelines for Writing Guidelines
+
+## Use this template as a guide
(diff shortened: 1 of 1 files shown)


Gudahtt (approved):
LGTM!

brad-decker (approved):

PR #11, opened 2022-10-11
Add initial guidelines for JavaScript and TypeScript

### LINKED ISSUE #10: Add initial JavaScript and TypeScript guidelines
We started a set of guidelines in Notion. We should copy them over.

mcmire:
These were copied from the notes that I've been keeping in Notion (see "Code Guidelines"). I tried to list best practices that I thought were already being followed in some form, so hopefully none of them should be too controversial.

Fixes #10.

[👉🏻 **See rendered JavaScript guidelines** 👈🏻](https://github.com/MetaMask/contributor-docs/blob/add-js-and-ts-guidelines/guides/javascript.md)
[👉🏻 **See rendered TypeScript guidelines** 👈🏻](https://github.com/MetaMask/contributor-docs/blob/add-js-and-ts-guidelines/guides/typescript.md)

### DIFF
--- a/guides/javascript.md
+++ b/guides/javascript.md
@@ -251,7 +251,7 @@ Jest incorporates most of the features of Sinon with a slimmer API:
 * `jest.fn()` can be used in place of `sinon.stub()`.
 * `jest.spyOn(object, method) can be used in place of `sinon.spy(object, method)`.
 * `jest.spyOn(object, method)` can be used in place of `sinon.stub(object, method)` (with the caveat that the method being spied upon will still be called by default).
-* `jest.useFakeTimers()` can be used in place of `sinon.useFakeTimers()` (though note that Jest's "clock" object has a different API than Sinon's and offers fewer options for advancing time).
+* `jest.useFakeTimers()` can be used in place of `sinon.useFakeTimers()` (though note that Jest's "clock" object had fewer features than Sinon's prior to Jest v29.5).
 
 ## Don't test private code
 
--- a/guides/javascript.md
+++ b/guides/javascript.md
@@ -149,7 +149,7 @@ Asynchronous code written using `async`/`await` syntax looks less complex and mo
 
 ``` javascript
 function makeRequest() {
-  return fetch("https://google.com")
+  return fetch('https://google.com')
     .then((response) => {
       return response.json().then((json) => {
         return json['page_views'];
--- a/guides/javascript.md
+++ b/guides/javascript.md
@@ -103,15 +103,7 @@ const [isRemoved, setIsRemoved] = useState(false);
 
 ## 💡 Place names of secondary concepts first in boolean variable names
 
-Sometimes a boolean variable describes a subject that's different than the context where the variable is defined:
-
-``` javascript
-const ownedAccount = Boolean(ownedAccountName);
-```
-
-In naming such a variable, there are a couple of different routes you could take.
-
-One way is to always place the verb at the beginning:
+When naming a boolean variable that includes a subject, the previous guideline suggests that you can place the verb at the beginning:
 
 ``` javascript
 const isRecipientOwnedAccount = Boolean(ownedAccountName);
--- a/guides/javascript.md
+++ b/guides/javascript.md
@@ -162,7 +162,7 @@ function makeRequest() {
 
 ``` javascript
 async function makeRequest() {
-  const response = await fetch("https://google.com");
+  const response = await fetch('https://google.com');
   const json = await response.json();
   return json['page_views'];
 }
--- a/guides/javascript.md
+++ b/guides/javascript.md
@@ -1,0 +1,215 @@
+# JavaScript
+
+## Use the latest version of our ESLint config
+
+If you're working on a project, there's a good chance that you're using our shared ESLint rules, which codifies decisions we've made around code style and can be used to align that project with other projects in our ecosystem. Make sure the project is up to date with the various packages within the [`eslint-config`](https://github.com/MetaMask/eslint-config) monorepo.
+
+## Use a verb phrase for function/method names
+
+When naming a function or method, describe the action, not the outcome.
+
+🚫
+
+``` javascript
+function formattedChangelog() {
+  // ...
+}
+```
+
+✅
+
+``` javascript
+function formatChangelog() {
+  // ...
+}
+```
+
+For a function or method that returns a boolean, prefix the name with a descriptive verb so that readers don't confuse it with a boolean variable.
+
+🚫
+
+``` javascript
+class NetworkController {
+  isEIP1559Compatible() {
+    // ...
+  }
+}
+```
+
+✅
+
+``` javascript
+class NetworkController {
+  determineEIP1559Compatibility() {
+    // ...
+  }
+}
+```
+
+### Read more
+
+- [Google's JavaScript Style Guide](https://google.github.io/styleguide/jsguide.html#naming-method-names)
+- [.NET Fundamentals](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-methods)
+- [Kotlin Coding Conventions](https://kotlinlang.org/docs/coding-conventions.html#choose-good-names)
+- [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/#strive-for-fluent-usage)
+
+## Use a verb phrase for boolean variable names that do not involve secondary objects
+
+Sometimes a variable that is intended to hold a boolean value does not have an explicit subject, but represents the context where the variable is defined (e.g., a class or an entire file). When naming such a variable, use a statement that describes the state of the subject, minus the name of the subject itself. Usually this means prefixing the name with a form of "to be" or "to have" (e.g. `is*`, `has*`), but you may find it more readable to use past or future tense and/or a modal verb such as `should`.
+
+🚫
+
+``` javascript
+const removed = false;
+```
+
+✅
+
+``` javascript
+// Any of these would do
+const isRemoved = false;
+const wasRemoved = false;
+const hasBeenRemoved = false;
+const shouldBeRemoved = false;
+```
+
+If the name represents a negative statement, reword it into a positive statement and inverting the value assigned to the value.
+
+🚫
+
+``` javascript
+const notEnoughGas = false;
+```
+
+✅
+
+``` javascript
+const hasEnoughGas = true;
+```
+
+Take special note of variables which are created via React's `useState` hook.
+
+🚫
+
+``` javascript
+const [removed, setRemoved] = useState(false);
+```
+
+✅
+
+``` javascript
+const [isRemoved, setIsRemoved] = useState(false);
+```
+
+## 💡 Place names of secondary concepts first in boolean variable names
+
+Sometimes a boolean variable describes a subject that's different than the context where the variable is defined:
+
+``` javascript
+const ownedAccount = Boolean(ownedAccountName);
+```
+
+In naming such a variable, there are a couple of different routes you could take.
+
+One way is to always place the verb at the beginning:
+
+``` javascript
+const isRecipientOwnedAccount = Boolean(ownedAccountName);
+```
+
+However, this naming strategy creates a point of friction for objects, arrays, or React components, where it may be desirable to sort identifiers alphabetically. In that case you could end up with something like:
+
+``` jsx
+<SenderToRecipient
+  isRecipientOwnedAccount={isRecipientOwnedAccount}
+  recipientName={toName}
+  recipientNickname={toNickname}
+  senderName={fromName}
+  senderAddress={fromAddress}
+/>
+```
+
+It is potentially more "greppable" if properties that concern the same concept are kept together instead of separate. To address this, you may wish to place the subject of the variable name at the beginning:
+
+``` javascript
+const recipientIsOwnedAccount = Boolean(ownedAccountName);
+```
+
+This would result in:
+
+``` jsx
+<SenderToRecipient
+  recipientIsOwnedAccount={isRecipientOwnedAccount}
+  recipientName={toName}
+  recipientNickname={toNickname}
+  senderName={fromName}
+  senderAddress={fromAddress}
+/>
+```
+
+## Use `async`/`await` syntax over `.then`/`.catch`
+
+Asynchronous code written using `async`/`await` syntax looks less complex and more straightforward than code written using `.then`/`.catch`. Additionally, using `async`/`await` leads to [better stack traces in Node and Chromium](https://mathiasbynens.be/notes/async-stack-traces), both of which use the V8 JavaScript engine, because when an asynchronous operation is `await`ed, the engine will remember the function where the `await` occurred, which means that it can place that function on the stack trace (otherwise, using `.then`/`.catch`, it would get lost).
+
+🚫
+
+``` javascript
+function makeRequest() {
+  return fetch("https://google.com")
+    .then((response) => {
+      return response.json().then((json) => {
+        return json['page_views'];
+      })
+    })
+}
+```
+
+✅
+
+``` javascript
+async function makeRequest() {
+  const response = await fetch("https://google.com");
+  const json = await response.json();
+  return json['page_views'];
+}
+```
+
+## `await` promises before `return`ing them
+
+Using `return await` rather than just `return` for an asynchronous operation leads to a more complete stack trace in Node and in Chromium browsers, both of which use the V8 JavaScript engine, because when an asynchronous operation is `await`ed, the engine will remember the function where the `await` occurred, which means that it can place that function on the stack trace (otherwise it would get lost).
+
+🚫
+
+``` javascript
+async function makeRequest() {
+  const response = await fetch('https://some/url');
+  return response.json();
+}
+```
+
+✅
+
+``` javascript
+async function makeRequest() {
+  const response = await fetch('https://some/url');
+  return await response.json();
+}
+```
+
+### Read more
+
+* <https://v8.dev/blog/fast-async#improved-developer-experience>
+* <https://docs.google.com/document/d/13Sy_kBIJGP0XT34V1CV3nkWya4TwYx9L3Yv45LdGB6Q/edit>
+
+## Use Jest's mock functions instead of Sinon
+
+Jest incorporates most of the features of Sinon with a slimmer API:
+
+* `jest.fn()` can be used in place of `sinon.stub()`.
+* `jest.spyOn(object, method) can be used in place of `sinon.spy(object, method)`.
+* `jest.spyOn(object, method)` can be used in place of `sinon.stub(object, method)` (with the caveat that the method being spied upon will still be called by default).
+* `jest.useFakeTimers()` can be used in place of `sinon.useFakeTimers()` (though note that Jest's "clock" object has a different API than Sinon's and offers fewer options for advancing time).
+
+## Don't test private code
+
+If you encounter a function or method whose name starts with `\_` or which has been tagged with the `@private` JSDoc tag or `private` keyword in TypeScript, you do not need to test this code. These markers are used to indicate that the code in question should not be used outside the file. As a result, any tests written for this code would need to have knowledge that no other part of the system would have. This is not to say that private code should not be tested at all, but rather that it should be tested via public code that makes use of the private code.
--- a/guides/javascript.md
+++ b/guides/javascript.md
@@ -174,9 +174,60 @@ async function makeRequest() {
 }
 ```
 
-## `await` promises before `return`ing them
+## `await` promises before returning them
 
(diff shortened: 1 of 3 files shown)


guides/javascript.md
```diff
@@ -0,0 +1,176 @@
+# JavaScript
+
+**NOTE:** All of these guidelines apply to TypeScript as well.
```
mcmire, line 3: So there are some guidelines which apply to JavaScript and TypeScript, and some which apply only to TypeScript. I wasn't sure how best to represent this. I figured that anything applicable to JavaScript was also applicable to TypeScript (but not vice versa) but I'm not sure if that's immediately obvious or not.

guides/javascript.md
```diff
@@ -0,0 +1,176 @@
...
+  selectedAddressSubjectMap && selectedAddressSubjectMap[originOfCurrentTab]
+);
+```
+
+say this:
+
+``` javascript
+const [isRemoved, setIsRemoved] = useState(false);
+const isSendDisabled = standard !== ERC721;
+const hasEnoughGas = isBalanceSufficient(...);
```
brad-decker, line 83: Not worth holding this up, but just wanna throw out something. For objects we'd like to enforce alphabetical assortment of keys, and sometimes that results in ordering that is not easily grepped for logic. Example:
```js
const state = {
  hasEnoughGas: false,
  isRecipientCopied: false,
  isGasSet: false,
};
```

in the above example, there is a unrelated state key injected between two quasi-related state keys.

```js
const state = {
  gasIsSet: false,
  gasCostIsCovered: false,
  recipientIsCopied: false,
};
```

The above example is off the top of my head, but the idea here is to put the noun/primary object first. 

Mainly commenting here to get your thoughts, not to suggest we change this. The written content already reflects an improvement in a lot of areas of our legacy code.

mcmire: Okay, I can get behind that. I know that you also proposed something similar here, is this related? https://github.com/MetaMask/metamask-extension/pull/14784#discussion_r882774247

brad-decker: I just realized that my original comment had incorrect ordering lol and wouldn't have been a problem, but i'll leave it. This is why we need linting for this rule :P 

Yeah thats exactly what triggered this comment.

brad-decker: Also if the noun first paradigm costs more overhead than the object values being not grouped together then we can avoid this convention entirely. Just one of those things that irks me that may be just a personal thing and not widely felt.

Gudahtt: I like the idea, but enforcing it would be challenging.

I'm reminded here of the Redux Style Guide, which has separate sections for "Essentials", "Strongly Recommended", and "Recommended".

Noun-first naming seems like a good candidate for the "Recommended" category.

I was considering creating a page on naming things; maybe I will try to fit this in that document.

brad-decker: I like the idea of having those same categories for our best practices, and i'm fine with the noun first naming being "recommended"

mcmire: Okay, I've added a new guideline which incorporates your suggestion here: [f9caede](https://github.com/MetaMask/contributor-docs/pull/11/commits/f9caede572b83f19ec6da5570db83b611fc1d49b). However this includes a bunch of other changes, notably the addition of guideline categories, which I've extracted to this PR: https://github.com/MetaMask/contributor-docs/pull/12

guides/javascript.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+### Read more
+
+- [Google's JavaScript Style Guide](https://google.github.io/styleguide/jsguide.html#naming-method-names)
+- [.NET Fundamentals](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-methods)
+- [Kotlin Coding Conventions](https://kotlinlang.org/docs/coding-conventions.html#choose-good-names)
+- [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/#strive-for-fluent-usage)
+
+## Use a verb phrase for boolean variable names that do not involve secondary objects
```
Mrtenz, line 56: This can be enforced with ESLint.

guides/javascript.md
```diff
@@ -0,0 +1,215 @@
...
+
+``` javascript
+async function makeRequest() {
+  const response = await fetch("https://google.com");
+  const json = await response.json();
+  return json['page_views'];
+}
+```
+
+## `await` promises before `return`ing them
```
Mrtenz, line 177: I agree with this, but it should be noted that this makes code significantly slower:

https://arthur.place/the-cost-of-return-await

guides/javascript.md
mcmire: Update: I ended up marking this guideline with a 💡 to indicate that it's more of a suggestion with alternatives.

guides/javascript.md
mcmire: Good point! I can add a disclaimer.

guides/typescript.md
```diff
@@ -0,0 +1,41 @@
+# TypeScript
+
+These guidelines specifically apply to TypeScript.
+
+## Provide explicit return types for functions/methods
```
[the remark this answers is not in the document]
mcmire: Good point! I've added the [`explicit-function-return-type`](https://typescript-eslint.io/rules/explicit-function-return-type/) rule here: https://github.com/MetaMask/eslint-config/pull/314

guides/javascript.md
mcmire: This guideline would be a bit tricky to enforce, I feel, because of the guideline after this, which suggests that boolean variable names may not be required to start with `is`, `has`, etc., if they involve objects secondary to the context. A project that wanted to use this strategy would then have to disable a rule we added for boolean variable names, as that rule would have to be generic. Would we want that?

guides/javascript.md
mcmire: It looks like this can also be enforced with the [`return-await`](https://typescript-eslint.io/rules/return-await/) rule.

guides/typescript.md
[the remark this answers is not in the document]
mcmire: By the way, given that we have some ESLint rules that we've enabled ourselves, what do we feel about explicitly documenting the reason for doing so? That means we'd keep this here but then have something at the top that says something like:

``` markdown
> **Note**  
> This guideline is enforced via [`explicit-function-return-type`](/link/to/rule/in/our/eslint-config).
```

guides/javascript.md
mcmire: By the way, for posterity, I read the article you linked to and came to [this bit at the end](https://arthur.place/the-cost-of-return-await#zero-cost-async-stack-traces) which suggests that the benefit we think `await` is providing here may not be necessary, but I don't think that's true. I am going to add an example to this section to prove that `await` is still necessary to preserve async functions in backtraces. I think [zero-cost async stack traces](https://docs.google.com/document/d/13Sy_kBIJGP0XT34V1CV3nkWya4TwYx9L3Yv45LdGB6Q/edit) is the thing that makes this even possible, but [that was added way back in v8 7.3](https://v8.dev/blog/v8-release-73#async-stack-traces) (or [Node 12, which used v7 7.8](https://nodejs.org/en/download/releases)), so I don't think we need to even mention that.

Gudahtt: Here's the example I used last time in a Slack discussion about this: https://consensys.slack.com/archives/G1L7H42BT/p1646953758233329?thread_ts=1646952599.353159&cid=G1L7H42BT

guides/javascript.md
```diff
@@ -0,0 +1,215 @@
...
+    // ...
+  }
+}
+```
+
+✅
+
+``` javascript
+class NetworkController {
+  determineEIP1559Compatibility() {
```
Gudahtt, line 43: Hmm, not sure how I feel about this one. I like how concise `is[condition]` is.

I struggle to come up with non-is names for these cases too, particularly for the verb part. e.g. here it could be "determine" or 'calculate" or 'derive" or "check" or "verify" or "validate", etc. It's like a filler word.

Regarding the negative consequence:

>so that readers don't confuse it with a boolean variable

For me personally this is not a relatable experience. Maybe it's because of my editor highlighting, or because we tend to work more with TypeScript these days, but I can't recall ever confusing an `is___` function for a boolean variable.

Though... I do remember running into a similar problem. When we started introducing the ESLint `no-shadow` rule, I had to rename a bunch of `is___` variables to avoid conflict with `is___` functions.

Gudahtt: I couldn't find any examples of this guideline in the referenced style guides either

mcmire: > I do remember running into a similar problem. When we started introducing the ESLint `no-shadow` rule, I had to rename a bunch of `is___` variables to avoid conflict with `is___` functions.

Ah. I have run into this too. I wonder if that's a better rationale.

guides/javascript.md
mcmire: I made some updates here: https://github.com/MetaMask/contributor-docs/compare/acbcc47...d632a27

Gudahtt: Hah, I just noticed that the referenced negative consequences don't apply to us anyway. We haven't transpiled async/await [since mid-2019](https://github.com/MetaMask/metamask-extension/pull/6812). The benchmarks shown for native async/await look pretty decent.

guides/javascript.md
```diff
@@ -0,0 +1,264 @@
...
+
+✅
+
+``` javascript
+const [isRemoved, setIsRemoved] = useState(false);
+```
+
+## 💡 Place names of secondary concepts first in boolean variable names
+
+Sometimes a boolean variable describes a subject that's different than the context where the variable is defined:
```
Gudahtt, line 106: This example was a little hard to understand for me. I'm not sure what the subject or the context are supposed to be here.

However, the later examples seemed very clear and easy to understand. Maybe this would be simpler if we got to these examples sooner? e.g. we could skip straight to the `isRecipientOwnedAccount` example, skipping these first two.

guides/javascript.md
```diff
@@ -0,0 +1,264 @@
...
+
+``` javascript
+const ownedAccount = Boolean(ownedAccountName);
+```
+
+In naming such a variable, there are a couple of different routes you could take.
+
+One way is to always place the verb at the beginning:
```
Gudahtt, line 114: This example was a little hard to understand for me. I'm not sure what the subject or the context are supposed to be here.

However, the later examples seemed very clear and easy to understand. Maybe this would be simpler if we got to these examples sooner? e.g. we could skip straight to the `isRecipientOwnedAccount` example, skipping these first two.

Something like:

```suggestion
When naming a boolean variable that includes a subject, the previous guideline suggests that you can place the verb at the beginning:
```

Gudahtt: Hah, just noticed that we're calling "is" a verb here. That doesn't seem right.

guides/javascript.md
```diff
@@ -0,0 +1,264 @@
...
+
+``` javascript
+const isRecipientOwnedAccount = Boolean(ownedAccountName);
+```
+
+However, this naming strategy creates a point of friction for objects, arrays, or React components, where it may be desirable to sort identifiers alphabetically. In that case you could end up with something like:
+
+``` jsx
+<SenderToRecipient
+  isRecipientOwnedAccount={isRecipientOwnedAccount}
```
Gudahtt, line 124: This example might be improved if we could think of a prop to go in-between this one as the `recipient__` ones, Just to make super clear what the advantages are of this naming convention.

Gudahtt: onClick maybe?

guides/javascript.md
```diff
@@ -0,0 +1,264 @@
...
+> `await`ing promises has a significant performance impact when transpiling to ES6. [**Read more**](https://arthur.place/the-cost-of-return-await).
+
+## Use Jest's mock functions instead of Sinon
+
+Jest incorporates most of the features of Sinon with a slimmer API:
+
+* `jest.fn()` can be used in place of `sinon.stub()`.
+* `jest.spyOn(object, method) can be used in place of `sinon.spy(object, method)`.
+* `jest.spyOn(object, method)` can be used in place of `sinon.stub(object, method)` (with the caveat that the method being spied upon will still be called by default).
+* `jest.useFakeTimers()` can be used in place of `sinon.useFakeTimers()` (though note that Jest's "clock" object has a different API than Sinon's and offers fewer options for advancing time).
```
Gudahtt, line 260: As of Jest v29.5, they might finally have all of the Sinon fake timer API methods exposed (see https://jestjs.io/docs/jest-object#jestadvancetimersbytimeasyncmstorun)

I was about to comment that maybe we shouldn't recommend Jest fake timers for now, but I guess it should be safe now. Maybe instead of this disclaimer about Jest having a different API with fewer options, we can add a note about the API missing features in versions prior to v29.5? e.g.

```suggestion
* `jest.useFakeTimers()` can be used in place of `sinon.useFakeTimers()` (though note that Jest's "clock" object had fewer features than Sinon's prior to Jest v29.5).
```

guides/javascript.md
```diff
@@ -0,0 +1,264 @@
...
+async function makeRequest() {
+  const response = await fetch('https://some/url');
+  return await response.json();
+}
+```
+
+> [!WARNING]\
+> `await`ing promises has a significant performance impact when transpiling to ES6. [**Read more**](https://arthur.place/the-cost-of-return-await).
+
+## Use Jest's mock functions instead of Sinon
```
Gudahtt, line 253: These two sections seem out-of-place in this article. They seem better suited to an article about unit tests.

guides/typescript.md
Gudahtt, line 1: Things that would be nice to have guidelines for:

* prefer `ts-expect-error` over `ts-ignore`
* When should we use `ts-expect-error`?
* When is it OK to use `any`?
* How to effectively narrow types (particularly when dealing with `unknown`).
  * Related: how to write type guard functions

guides/javascript.md
mcmire: Okay, so it sounds like we want to remove this disclaimer then?

guides/javascript.md
mcmire: I've made the requested change here: [6d93d07](https://github.com/MetaMask/contributor-docs/pull/11/commits/6d93d07187b2821d5d9de499bd10c77a714efb21)

guides/javascript.md
mcmire: I also wasn't quite sure, but I remember looking this up before typing it. But I've looked it up again to be double-sure.

It looks like "is" is a verb, but it's special. It doesn't do anything but tie the subject and object together, so sometimes it's called a "linking" or "copula" verb: https://en.wikipedia.org/wiki/Linking_verb

guides/javascript.md
mcmire: Makes sense, updated here: [7b8e6ae](https://github.com/MetaMask/contributor-docs/pull/11/commits/7b8e6ae0de276d38b6e58adae5a97d3140738ed9)

guides/javascript.md
mcmire: Yeah, I think I'd separated testing guidelines in the original Notion doc, so that makes sense. I'll extract it for now, and we can expand on it in a future PR.

mcmire: Extracted here: [aa8fa2f](https://github.com/MetaMask/contributor-docs/pull/11/commits/aa8fa2f9319ca1140e556aa8fe4c3c512e5e5049)

guides/javascript.md
mcmire: Never mind, I see both thumbs-ups. Disclaimer removed here: [22a9daa](https://github.com/MetaMask/contributor-docs/pull/11/commits/22a9daa2143c7632f13018ae5431ac950357af66)

guides/typescript.md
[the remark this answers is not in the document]
Gudahtt: That seems like a good idea to me. We can remove it later if we don't think it has been useful.

Generally it would be nice to keep this guide shorter, and let ESLint explain things for us where possible. But the ESLint rules don't explain our reasoning very well. I wish we could.... explain the reasons for lint rules inline and advice a course of action, something like that. Ah well.

guides/testing.md
```diff
@@ -0,0 +1,14 @@
...
+## Use Jest's mock functions instead of Sinon
+
+Jest incorporates most of the features of Sinon with a slimmer API:
+
+* `jest.fn()` can be used in place of `sinon.stub()`.
+* `jest.spyOn(object, method) can be used in place of `sinon.spy(object, method)`.
+* `jest.spyOn(object, method)` can be used in place of `sinon.stub(object, method)` (with the caveat that the method being spied upon will still be called by default).
+* `jest.useFakeTimers()` can be used in place of `sinon.useFakeTimers()` (though note that Jest's "clock" object had fewer features than Sinon's prior to Jest v29.5).
+
+## Don't test private code
```
Gudahtt, line 12: It would be nice to expand more on the reasoning behind this.
e.g.
* Explain how this ties in with a broader BDD-like testing philosophy,
* Tests are more useful for preventing regressions and helping document intended behavior when they only reflect the public API
* This ensures the tests help make refactoring easier, rather than getting in the way

This is a great start though! I can follow up to add more detail later.

guides/javascript.md
```diff
@@ -0,0 +1,242 @@
...
+  isRecipientOwnedAccount={isRecipientOwnedAccount}
+  onClick={onClick}
+  recipientName={toName}
+  recipientNickname={toNickname}
+  senderAddress={fromAddress}
+  senderName={fromName}
+/>
+```
+
+It is potentially more "greppable" if properties that concern the same concept are kept together instead of separate. To address this, you may wish to place the subject of the variable name at the beginning:
```
Gudahtt, line 125: Nit: This might me more readable (:sweat_smile: ) if we say this is easier to read, rather than saying "greppable". I'm not sure it really helps with code searching, given that you could do a case-insensitive search for "recipient" and still find it either way. But the readability difference is clear

```suggestion
It is potentially easier to read if properties that concern the same concept are kept together instead of separate. To address this, you may wish to place the subject of the variable name at the beginning:
```

Gudahtt (dismissed):
LGTM!

guides/typescript.md
mcmire: Okay! I will add these in a followup PR.

guides/typescript.md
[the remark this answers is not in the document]
mcmire: Agreed... okay. I can add this in a followup PR once we have a new version of `eslint-config` out.

guides/testing.md
mcmire: Sounds good!

guides/javascript.md
mcmire: In using "greppable" I borrowed @brad-decker's wording from the [discussion above](https://github.com/MetaMask/contributor-docs/pull/11#discussion_r995999516). After reading it again, though, I wonder if he might have meant "scannable"/"readable" instead. So I like this suggestion, but considering that you've already approved I will make this change in a future PR (pending possible confirmation).

guides/javascript.md
mcmire: Oops I didn't mean to resolve this but I hope that makes sense ^

Gudahtt: Yep, makes sense!

guides/javascript.md
```diff
@@ -0,0 +1,242 @@
...
+
+## Use `async`/`await` syntax over `.then`/`.catch`
+
+Asynchronous code written using `async`/`await` syntax looks less complex and more straightforward than code written using `.then`/`.catch`. Additionally, using `async`/`await` leads to [better stack traces in Node and Chromium](https://mathiasbynens.be/notes/async-stack-traces), both of which use the V8 JavaScript engine, because when an asynchronous operation is `await`ed, the engine will remember the function where the `await` occurred, which means that it can place that function on the stack trace (otherwise, using `.then`/`.catch`, it would get lost).
+
+🚫
+
+``` javascript
+function makeRequest() {
+  return fetch("https://google.com")
```
Mrtenz, line 152: Nit: We use single quotes everywhere.
```suggestion
  return fetch('https://google.com')
```

guides/javascript.md
```diff
@@ -0,0 +1,242 @@
...
+      })
+    })
+}
+```
+
+✅
+
+``` javascript
+async function makeRequest() {
+  const response = await fetch("https://google.com");
```
Mrtenz, line 165: Nit: We use single quotes everywhere.
```suggestion
  const response = await fetch('https://google.com');
```

Mrtenz (changes requested):

guides/javascript.md
mcmire: Good point! I need to add Prettier to this repo with the same settings that we usually use. I can fix this for now though.

guides/javascript.md
mcmire: I had to push up another commit anyway so I added your suggestion.

Gudahtt (approved):

Mrtenz (approved):

brad-decker (approved):
⭐ 💯

PR #12, opened 2022-11-12
Add guideline categories

mcmire:
Some guidelines are more important to follow than others; some guidelines have alternatives which may be preferable depending on the circumstance. To clarify how guidelines ought to be followed, define importance levels/categories, and assign each guideline to one in the "meta" guidelines document as an example. These categories are borrowed from the [Redux Style Guide](https://redux.js.org/style-guide/#rule-categories). To prevent from having to create separate category sections in each guidelines document, I've opted to associate an emoji with each category.

### DIFF
--- a/guides/guidelines.md
+++ b/guides/guidelines.md
@@ -1,18 +1,26 @@
 # Guidelines for Writing Guidelines
 
-## Use this template as a guide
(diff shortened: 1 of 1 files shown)


Gudahtt (approved):
LGTM!

PR #33, opened 2023-07-18
Update guidelines for guidelines

mcmire:
There are a couple of issues with the "meta-guidelines":

1. Currently, if you want to assign a guideline to a category, you must
   use an emoji in the guideline header. There are two problems with
   emoji: they are difficult to work with — try selecting one in your
   editor! — and, like icons, they can mean different things to
   different people.

2. Originally we borrowed the concept of guideline categories from the
   Redux Style Guide. This guide lists three of them: Essential,
   Strongly Recommended (renamed to "Recommended" here), and Recommended
   (renamed to "Suggested"). However, there is not a clear delineation
   between these categories — when do you know whether to assign
   something as Recommended vs. Essential? — and anyway, we've only used
   the first and third categories so far.

3. The guideline that describes how to present a dichotomy — a
   non-recommended way to do something and a recommended way to do it —
   allowed for unclear wording.

4. The "be empowered" guideline isn't really a guideline, it's just a
   nice thing to say. And it really applies to the whole repo, not just
   the guidelines.

To solve these issues:

1. I've simplified the concept of guideline categories. There are
   only two. All guidelines are essentially rules by default. If a
   guideline presents options for addressing a pain point, they can
   be called out by using an emoji.

2. I've simplified the instructions on how to give examples.

3. I've removed the "be empowered" guideline. I may add that to the root
   README and also explain how to propose new changes.

**[👉🏻 Rendered version 👈🏻 ](https://github.com/MetaMask/contributor-docs/blob/update-guidelines-for-guidelines/guides/guidelines.md)**

### DIFF
--- a/guides/guidelines.md
+++ b/guides/guidelines.md
@@ -1,77 +1,51 @@
 # Guidelines for Writing Guidelines
 
-## 🙏 Use this document as a guide
+## Use this document as a guide
 
 This document is not only a set of guidelines, but also a template for other guidelines documents.
 
-## 🙏 Communicate the importance of your guideline
-
-Guidelines may differ in how closely they are intended to be followed. To designate the importance of a guideline, include an emoji in its header to assign it to one of three categories:
-
-1. **🙏 (Standard):** These are guidelines that it really behooves everyone to follow. There shouldn't be a reason not to follow them, but anyone who feels that a guideline should be changed or removed should start a discussion in a pull request.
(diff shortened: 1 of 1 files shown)


mcmire:
Still workshopping this :)

mcmire:
This is now ready for review again.

Gudahtt (approved):
LGTM!

PR #37, opened 2023-08-01
Explain connection to ESLint rules

mcmire:
There is a notable overlap between the code guidelines in this repository and our ESLint rules. This talks about why guidelines are still useful in this context. It also removes the reference to "we" that slipped into a previous change.


Gudahtt (approved):
LGTM!

Issue #25, opened 2022-12-08
Standards: Unit testing


xinnanyemm:
This document would cover best practices and guidelines on writing tests that are complete and maintainable. Sections could include:
- The importance of naming tests well
- Organizing tests
- The four-phase test pattern
- Why reducing duplication in tests is an anti-goal
- Writing isolated tests (and why Jest hooks are counterproductive)

Gudahtt:
We recently discussed test naming conventions as a team, and have agreed that we should not prefix test descriptions with "should". We will use BDD-style descriptions, but they will assert what the code-under-test is doing rather than saying what it should do.

mcmire:
Note that this ticket is essentially a subset of https://github.com/MetaMask/contributor-docs/issues/5. For now we'll focus on unit testing and we can incorporate into a complete guide on testing later.

mcmire:
For context in estimation, I plan on adding more guidelines similar to those I've added for JavaScript and TypeScript. In other words, they will be essentially bullet points that act as guideposts for writing better tests, with a blurb for each point and some examples. I don't plan on writing anything that's long-form, despite what the description might indicate.

PR #38, opened 2023-08-15
Extend testing guidelines

mcmire:
This PR proposes a new set of guidelines for writing unit tests across our projects. Most are new, or if they are followed to some degree, they have not yet been codified.

Fixes https://github.com/MetaMask/contributor-docs/issues/25.

**[👉 View rendered document 👈](https://github.com/MetaMask/contributor-docs/blob/add-more-testing-guidelines/docs/unit-testing.md)**


guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+      });
+      await keyringController.submitPassword('abc123');
+
+      expect(keyringController.password).toBe('abc123');
+    });
+  });
+})
+```
+
+## Use Jest's mock functions instead of Sinon
```
mcmire, line 967: This is a corollary of "Use Jest", but I chose to call it out because we have a lot of older tests that were presumably written before the conversion to Jest and use Sinon.

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+    });
+
+    it('requests to unsubscribe if the block tracker is already stopped', () => {
+      // ...
+    });
+  });
+});
+```
+
+## Be conscious of phases
```
mcmire, line 341: At face value, this is less of a guideline and more of a sidebar. I wanted to call it out because I reference phases numerous times throughout this document. I do provide a suggestion for highlighting phases below, however, so _that_ is a guideline. I'm wondering whether it would make more sense to reword this guideline to fit the suggestion?

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+    // ...
+  });
+
+  it('returns a promise that resolves to true', () => {
+    // ...
+  });
+});
+```
+
+## Use `it` to specify known behavior for a unit of code
```
mcmire, line 163: Originally I had a section which explained the purpose of test names and why they're arranged using `describe` and `it`. Essentially, you can draw a straight line from Jest to the birth of behavior-driven development. However, BDD is a subject unto itself, and I didn't want to overload people with information.

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+  })
+});
+```
+
+### Read more
+
+- ["Arrange, act, assert"](http://xp123.com/articles/3a-arrange-act-assert/) is a simplification of the four-phase test.
+- In [behavior-driven development](https://www.agilealliance.org/glossary/bdd/), this is also called ["given, when, then"](https://martinfowler.com/bliki/GivenWhenThen.html).
+
+## Keep tests isolated
```
mcmire, line 423: This guideline is very long, and conceivably each section could be its own guideline, since there is different advice for different situations. It's also a little more philosophical than other guidelines. Let me know if anything is unclear or whether it makes sense to arrange this a different way.

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+});
+```
+
+## Use `it` to specify known behavior for a unit of code
+
+Use `it` to describe the behavior of a piece of code from the consumer's perspective as it stands today, not as it ought to be.
+
+Do not use "should" at the start of the test name. The official Jest documentation [omits this word from their examples](https://jestjs.io/docs/next/getting-started), and it creates noise when reviewing the list of tests that Jest outputs after a run.
+
+Do not repeat the name of the function or method in the name of the test if it is possible to communicate what a consumer expects more clearly.
```
mcmire, line 169: There's another bit of advice that could go here, which is to be aware of scenarios — changes in the context which affect the behavior of the thing under test — and call them out in the test name. That probably deserves its own guideline, however, and I've already written a lot 😅

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+
+## Use Jest
+
+[Jest](https://jestjs.io/) is the testing framework of choice for unit tests at MetaMask. It offers a more comprehensive set of features out of the box over Mocha or Tape that make writing tests for production code very easy. Some of the useful features are:
+
+- [Module mocks](https://jestjs.io/docs/next/mock-functions#mocking-modules)
+- [Timer mocks](https://jestjs.io/docs/next/timer-mocks)
+- [Snapshots](https://jestjs.io/docs/next/snapshot-testing)
+- Automatic [parallelization of tests](https://jestjs.io/blog/2016/09/01/jest-15#buffered-console-messages) without [restricting usage of exclusive tests](https://mochajs.org/#exclusive-tests-are-disallowed)
+- [Automatic sandboxing of tests](https://jestjs.io/blog/2016/03/11/javascript-unit-testing-performance) to prevent accidental leaks when modules are mocked
```
Gudahtt, line 11: Hmm... not sure about this one. I know Jest runs each test module in a "sandbox" but the individual tests within that module aren't isolated from each other automatically.

It's hard to tell what this linked post is referring to - maybe it's talking about how their virtualized environment allows effectively mocking `require` and resetting it between each test run? But even that doesn't happen automatically (AFAIK), you need to set `resetMocks` or manually call that between tests.

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+
+it('returns a promise that resolves to true if "addToken" is true', () => {
+  // ...
+})
+```
+
+2️⃣
+
+``` typescript
+describe('if "addToken" is true', () => {
```
Gudahtt, line 148: Nit: I think it would be better to encourage people to group tests by _scenario_, rather than by other similar factors. The description of this guideline kinda implies that already, but this example isn't as clear because this isn't written like a scenario.

Maybe a description like "when adding a token" would be better here? That might prompt people to think more about the purpose of the tests and the real scenarios they represent, rather than just on inputs and outputs.

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+
+  it('returns a promise that resolves to true', () => {
+    // ...
+  });
+});
+```
+
+## Use `it` to specify known behavior for a unit of code
+
+Use `it` to describe the behavior of a piece of code from the consumer's perspective as it stands today, not as it ought to be.
```
Gudahtt, line 165: Hmm. This doesn't seem right. I think test descriptions should describe intentions, i.e. what the code _should_ be doing. Of course they also have to be true (i.e. reflective of what the code actually does), but that's a given. Otherwise the tests wouldn't pass.

We could model the behaviour of the code in many different ways, but organizing by intent produces useful and readable tests.

guides/unit-testing.md
Gudahtt, line 169: Is the qualifier necessary here? e.g. can you think of an example where it would be difficult to avoid repeating the name of the function?
```suggestion
Do not repeat the name of the function or method in the name of the test.
```

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+  });
+})
+```
+
+✅
+
+``` typescript
+describe('TokensController', () => {
+  describe('addToken', () => {
+    it('adds the given token to the "tokens" array in state', () => {
```
Gudahtt, line 214: Nit: "tokens array" here seems like an implementaiton detail. Ideally the test descriptions would describe the behavior in high-level terms, so it's clear to the reader why we want this to happen, and so that we don't have to update the description if we refactor things (e.g. if we change the tokens array to a set, for example).
```suggestion
    it('adds the given token to state', () => {
```

guides/unit-testing.md
Gudahtt: Yeah, I agree. I was about to comment on the "All unit tests follow four phases" part, because that isn't really true (some tests repeat setup/exercise/verify steps). But we can say that we _recommend_ writing tests in these phases

guides/unit-testing.md
Gudahtt: The part about determinism maybe isn't strictly needed here. We can justify test isolation without talking about that at all. I tend to think about isolation more in terms of test maintainability and readability. e.g. 

* Tests should be isolated so that they don't behave differently when `.only` is used for debugging
* Tests should be isolated so that the setup steps fully describe the test scenario to readers
* Tests should be isolated so that they don't mysteriously break when changes are made to other tests.

Plus, that statement about determinism doesn't seem right anyway. Non-deterministic tests are super useful! See fuzz testing. I prefer such tests to not be included as PR checks, instead being run on some other internval (e.g. on push to `main`, nightly, something like that), but there's still a place for non-deterministic testing.

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+
+In order to be useful, tests must be deterministic. That is, a single test within a file must always pass each and every time, whether it is run on its own, alongside other tests in the same file, or other tests across other files.
+
+To achieve determinism, tests must be performed in a clean room. If a test makes changes to any kind of state which persists outside of itself, it must revert those changes after completion to prevent contaminating other tests.
+
+There are a few different forms that this state can take:
+
+### Module mocks
+
+Mocks for modules — objects reachable via `import` lines — are created via `jest.mock` declarations within a test file. Since modules are mocked once when a test file is read, they are global in the context of the file, so changes in one test can "bleed" over into another test:
```
Gudahtt, line 433: Nit: Seems a bit unnecessary to define "module" here - hopefully that's a commonly understood word (especially given your examples). Also it doesn't seem correct to call the `jest.mock` calls "declarations" - nothing is being declared when we mock a module.
```suggestion
Mocks for modules are created by calling `jest.mock` within a test file. Since modules are mocked once when a test file is read, they are global in the context of the file, so changes in one test can "bleed" over into another test:
```

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+    it("does something usually", () => {
+      // Now `config.isActive` will be `false` within `functionUnderTest`,
+      // so it may not do what you expect
+      expect(functionUnderTest()).toDoSomething();
+    })
+  });
+})
+```
+
+Setting Jest's [`resetModules`](https://jestjs.io/docs/configuration#resetmodules-boolean) option to `true` instructs Jest to recreate the module registry for each test (i.e. rerun the function that you pass to `jest.mock`), guaranteeing that state kept in modules gets reset appropriately. (This option is set in MetaMask's [module template](https://github.com/MetaMask/metamask-module-template).)
```
Gudahtt, line 467: Hmm. Technically we set `resetMocks` not `resetModules`, though mocks includes module mocks. The distinction seems unnecessary to explain here though. Maybe to avoid this confusion we can reference the `resetMocks` configuration option here instead?

Gudahtt: Actually.... does `resetMocks` include `resetModules`? I swear I've looked this up before but now I'm not sure. The docs are frustratingly ambiguous about this.

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+    expect(interpretMethodData(provider, '0x3fbac0ab')).toStrictEqual({
+      name: 'Set Approval For All'
+    });
+  });
+});
+```
+
+### Shared variables
+
+If a variable used inside of a test is declared outside of the test, the value that the variable holds — and thus changes to that variable made inside the test — will remain even after the test ends.
```
Gudahtt, line 643: Nit: I found the wording here to be a little confusing. I've attempted to re-word it while conveying the same information, do you think this sounds OK?
```suggestion
Variables declared outside of tests are not reset automatically between tests. Changes made to these variables in tests can affect later tests, breaking test isolation.
```

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+  it("returns the signature for setApprovalForAll on Mainnet", () => {
+    // Okay, this will still use Mainnet
+    expect(interpretMethodData('0x3fbac0ab', network)).toStrictEqual({
+      name: 'Set Approval For All'
+    });
+  });
+});
+```
+
+However, it is best to [avoid hooks altogether](#avoid-before-each-and-after-each). Instead, use a factory function which defines defaults and allows them to be overridden where they are needed:
```
Gudahtt, line 706: Nit: Slightly softer wording here might be better
```suggestion
However, it is recommended to [avoid hooks altogether](#avoid-before-each-and-after-each). Instead, use a factory function which defines defaults and allows them to be overridden where they are needed:
```

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+        },
+      });
+    });
+  });
+});
+```
+
+However, this strategy ends up increasing the long-term maintenance cost of tests for a couple of reasons:
+
+- It makes debugging tests difficult later. Data that is critical to the purpose of a test — i.e., that is used in the execution or verification phase — may be hidden in a `beforeEach` hook, or it may be split across `beforeEach` and the test in question. This means that in order to trace data as it moves through the test, one must spend time piecing together the full picture.
```
Gudahtt, line 838: Nit: What's described here sounds more like "reading" than "debugging". I have this problem not just when debugging, but also when reading a test to understand what it does, or reviewing a test in a PR.

```suggestion
- It makes tests harder to read. Data that is critical to the purpose of a test — i.e., that is used in the execution or verification phase — may be hidden in a `beforeEach` hook, or it may be split across `beforeEach` and the test in question. This means that in order to trace data as it moves through the test, one must spend time piecing together the full picture.
```

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+- It makes writing new tests difficult. Extracting the setup phase assumes that all tests will need to be set up in exactly the same way. But the very nature of code guarantees that this will not be the case. Inevitably, tests will involve scenarios that cause differences in behavior from the "happy path". Using `beforeEach`, in order to construct a new scenario, the test must extend what `beforeEach` is already doing, whether this is passing a different argument to the function being tested, mocking a function to return a different result, or something else. But this is not always as straightforward as it sounds. It is common to declare variables above the `beforeEach` which can be used later on in the file, but the utility of these variables is misleading, as by the time a test runs, these variables have already been used to set it up and thus cannot be reused or changed. Sometimes this creates enough of a problem that developers resort to copying the entirety of the setup code from `beforeEach` into the test, but this introduce another way that maintainers must now use to work with the tests.
```
Gudahtt, line 839: Nit: I get what you were going for here, but there's a lot going on here. I've tried to simplify it a bit. I found the examples later in this line to be too confusing as well - I think we'd need to give them more space if we want to keep them in.

```suggestion
- It makes writing tests for new scenarios difficult. The setup steps in `beforeEach` may not apply to future tests that cover different scenarios. In these cases, the author is forced to take on a complex refactor to remove the `beforeEach` steps that don't apply, or pursue complex workarounds.
```

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+
+A function solves the problems that hooks solve without introducing the problems that they simultaneously introduce. With this pattern, tests that need to customize setup data can do so easily by passing options to the function, and this draws a clear line from the verification phase to the setup phase:
```
Gudahtt, line 841: Nit: I've attempted to simplify the first sentence:

 ```suggestion
A setup function serves the same purpose without causing these problems. With this pattern, tests that need to customize setup data can do so easily by passing options to the function, and this draws a clear line from the verification phase to the setup phase:
```

File: guides/unit-testing.md [hunk 17]
```diff
@@ -0,0 +1,973 @@
...
+            symbol: 'DAI',
+          },
+        });
+      });
+    });
+  });
+});
+```
+
+The only case in which `beforeEach` or `afterEach` is acceptable is when using the Jest API, such as when setting fake timers:
```
Gudahtt, line 907: Nit: softer wording:
```suggestion
The only case in which `beforeEach` or `afterEach` is recommended is when using the Jest API, such as when setting fake timers:
```

Gudahtt: Actually.... is this recommended? `useFakeTimers` is similar to other setup steps, in that it might be useful for some steps but not others.

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+type WithControllerCallback = (setup: {
+  controller: TokensController;
+  onNetworkDidChangeListener: (networkState: NetworkState) => void;
+}) => void | Promise<void>;
+
+type WithControllerArgs =
+  | [WithControllerOptions, WithControllerCallback]
+  | [WithControllerCallback];
+
+async function withController(...args: WithControllerArgs) {
```
Gudahtt, line 857: It's great that we're showing an example here of the `with___` pattern, which is super useful for setting up things that also need a "teardown" phase. But this seems like the more complex case. I more often find myself in need of setup without any teardown, which is a lot simpler to write and demonstrate.

Maybe we could include both examples, and explain that this is recommended when there are teardown steps?

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+    },
+  }, fn] = args.length === 1 ? [{}, args[0]] : args;
+  const onNetworkDidChangeListener: (networkState: NetworkState) => void;
+
+  const controller = new TokensController({ chainId, onNetworkDidChange });
+  assert(onNetworkDidChangeListener, "onNetworkDidChangeListener was not set")
+
+  try {
+    await fn({ controller, onNetworkDidChangeListener });
+  } catch {
```
Gudahtt, line 871: Presumably we'll want to destroy the controller in all cases:
```suggestion
  } finally {
```

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+
+``` typescript
+const PASSWORD = 'abc123';
+
+// ... many lines later ...
+
+describe('KeyringController', () => {
+  describe('submitPassword', () => {
+    it('sets the password on the controller', async () => {
+      const keyringController = await initializeKeyringController({
```
Gudahtt, line 939: Hmm. This example doesn't look objectionable to me. The contents of the password aren't really necessary to read this test. As long as you assume it's a non-emtpy string, this is perfectly readable.

Did you intend to omit the password from the initialization options? That would be a much better example. Then we could add a comment explaining that it's not clear to the reader that the password was correct

Gudahtt (commented):
Great work! This is a really comprehensive guide.

There are a few high-level guidelines that I'd like to add at some point, but they can come later (e.g. what should you test, test coverage, refactor code-under-test to be more easily testable, etc.)

guides/unit-testing.md
mcmire: Hmm, yes, that link isn't very useful. I might have meant something else.

In any case, you're right that only test files are isolated, not tests themselves. There is probably some reason why the Jest team decided to add sandboxing but it seems like an implementation detail. I can remove this.

mcmire: Removed here: [18feed3](https://github.com/MetaMask/contributor-docs/pull/38/commits/18feed3683747443f588f8d6a62a1908602507fe)

guides/unit-testing.md
mcmire: I've made some changes to this guideline here: [b23c47](https://github.com/MetaMask/contributor-docs/pull/38/commits/7b23c477f506c5ae1f2bb4ca0a96d89cab8d8b0d)

guides/unit-testing.md
mcmire: Hmm. I see what you're saying. I'll work on this some more.

guides/unit-testing.md
mcmire: Oh, interesting. So according to the documentation:

- `resetMocks` calls `jest.resetAllMocks()` before each test. `jest.resetAllMocks()` calls `.mockReset()` on every mocked function.
- `resetModules` calls `jest.resetModules()` before each test. `jest.resetModules()` resets the module registry.
- `restoreMocks` calls `jest.restoreAllMocks()` before each test. `jest.restoreAllMocks()` calls `.mockRestore()` on every mocked function and `.restore()` on every replaced property. It only works on mocks created via `jest.spyOn()` and properties created with `jest.replaceProperty()`.

But according to the comments in `jest.config.js`:

- `resetMocks` resets all mocks, including mocked modules, to `jest.fn()` between each test case.
- `resetModules` resets the module registry before running each individual test.
- `restoreMocks` restores all mocks created using `jest.spyOn()` to their original implementation between each test. It does not affect mocked modules.

So... I'm unsure. According to the source code, [`jest.resetAllMocks()` calls `resetAllMocks` on a ModuleMocker instance](https://github.com/jestjs/jest/blob/331ad059334ab024d0f395cb49fd220d4bebd805/packages/jest-runtime/src/index.ts#L1351), but [`jest.resetModules()` does a bunch of stuff to reset the module registry](https://github.com/jestjs/jest/blob/331ad059334ab024d0f395cb49fd220d4bebd805/packages/jest-runtime/src/index.ts#L1197). So I am inclined to think that the comment for `resetAllMocks` is lying and that it _doesn't_ actually handle modules. But I'll try it out.

mcmire: After playing around with `resetModules`, it appears that it won't protect against the use case I presented in this "Module mocks" section. I think we would need to re-`require` module between each test if we really wanted to do that. That's a bit disappointing, but maybe I will experiment some more later. In the meantime I can remove this section.

mcmire: And your suggestion to recommend the use of the configuration option and put the details in a "Read more" section is good one, I can make that change.

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+
+## Keep tests isolated
+
+In order to be useful, tests must be deterministic. That is, a single test within a file must always pass each and every time, whether it is run on its own, alongside other tests in the same file, or other tests across other files.
+
+To achieve determinism, tests must be performed in a clean room. If a test makes changes to any kind of state which persists outside of itself, it must revert those changes after completion to prevent contaminating other tests.
+
+There are a few different forms that this state can take:
+
+### Module mocks
```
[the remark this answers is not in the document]
mcmire: Yeah, this is a good suggestion. I can make this change.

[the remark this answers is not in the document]
mcmire: Okay, I revised this section. Let me know what you think: [45403ef](https://github.com/MetaMask/contributor-docs/pull/38/commits/45403ef859a71d02a80596b6b6b6dfd810cec0c9)

guides/unit-testing.md
mcmire: Okay, I revised this section to recommend using `resetMocks` and `restoreMocks`: [45403ef](https://github.com/MetaMask/contributor-docs/pull/38/commits/45403ef859a71d02a80596b6b6b6dfd810cec0c9)

guides/unit-testing.md
mcmire: Good suggestion. I didn't see this before I revised this whole section, but I've applied the suggestion here: [48d3c7a](https://github.com/MetaMask/contributor-docs/pull/38/commits/48d3c7a481a227f9527ca03fc80eac0a6c6e404c)

guides/unit-testing.md
mcmire: In another subsection I used "Instead of using hooks, ..." so I copied that wording to here to be consistent. Does that phrase still need revision?

Gudahtt: That sounds good too! Just wanted to avoid "best"

guides/unit-testing.md
mcmire: You're right. I like your new wording, that sounds a lot better 😅 Changed here: [054f0c4](https://github.com/MetaMask/contributor-docs/pull/38/commits/054f0c43fa2488e6f92a7a6192970a21a3169029)

guides/unit-testing.md
mcmire: Thanks. Updated here: [079c957](https://github.com/MetaMask/contributor-docs/pull/38/commits/079c957dcf678dbaf0920c35af599a839faaa5c0)

guides/unit-testing.md
mcmire: Fair point. I included this exception because I've noticed that `useFakeTimers` tends to only be set once at the top of the file (because all of the tests rely on timers in some way) and I haven't run into a case where a single test needs something different. But, of course, it's always possible. In any case, this advice definitely conflicts with the advice given in the test isolation guidelines.

Should we recommend that people use a wrapper function instead?

``` typescript
function withFakeTimers(fn: () => void | Promise<void>) { 
  jest.useFakeTimers();

  try {
    await fn();
  } finally {
    jest.useRealTimers();
  }
}
```

guides/unit-testing.md
mcmire: Whoops! Yup, you're totally right. Fixed in [93daeb5](https://github.com/MetaMask/contributor-docs/pull/38/commits/93daeb5645adc7a3b39965b3fbb1237408de6142).

guides/unit-testing.md
mcmire: Yeah, definitely. I can add that.

guides/unit-testing.md
mcmire: I tweaked the wording some here, but I like your ideas for why test isolation exists. I'll see if I can work them in tomorrow.

guides/unit-testing.md
mcmire: By adding this clause I was trying to admit that naming is hard, and sometimes the method already describes what it does well and it's really hard to come up with other words. I guess that's not really what I'm saying though. I can remove this.

guides/unit-testing.md
mcmire: You're right. When I was writing this guideline I was thinking about where I got my current ideas about testing. I have some predispositions from working at consultancies and I remembered that there were two ideas floating around: 1) "tests as specifications" — which I interpret as "tests are valuable for verifying the behavior of a part of the system but also describing that behavior" — and 2) "tests as examples" — which I interpret as "tests are valuable for demonstrating to consumers how to use it".

At the same time, I was trying to think of a better reason for why it would be desirable to remove "should" from test descriptions other than "it's noise". That led me to the question, if tests describe the behavior of the system, and we don't want to use "should", what we are saying in the test descriptions? Are we describing how the system ought to work or how it actually works? I could see how it could be helpful to be honest and admit "here's what we've got today", whether it's desirable or not. This perspective is especially useful, I think, if you're backfilling tests for code that wasn't written with tests to begin with and you're discovering behavior that people may not realize is there.

That's what I was aiming for in writing this guideline, but I'm realizing that it's a nuanced viewpoint and not applicable in general situations.

All this to say, I agree with you that thinking about intended behavior is more important in most situations than thinking about the actual behavior.

So now the question is, if intention is important, do we want to be using "should" in tests? The [original thinking](https://dannorth.net/introducing-bdd/#an-expressive-test-name-is-helpful-when-a-test-fails) behind "should" was that it baked uncertainty into the software development process and served as a reminder to [constantly compare what the code is doing vs. what it ought to be doing (so, kind of the other side of the coin of what I said above). Is this valuable? Or do we already do this anyway?

Gudahtt: I do like your "tests as specifications" interpretation, that's how I think of them as well. But I would expect a specification to communicate the intention behind each guideline as well (at least for anything non-obvious).

There are some cases where we might be writing tests after-the-fact, and we find something non-desirable, and we want to model it with a unit test anyway just so we don't lose track of the behavior. But that would only happen when backfilling tests, so I think of that as an edge case. "Intention" maybe isn't as relevant if it's something a past author wrote that you don't agree with.

But if it's code you authored to serve a specific purpose, the test description is a great place to state that purpose.

Gudahtt: I'm still good with omitting "should". I saw it as being better to exclude because the tests then can be read as a specification, and also because it added verbosity to the test descriptions. But I never felt strongly about that guideline either way.

guides/unit-testing.md
Gudahtt: Nice!

guides/unit-testing.md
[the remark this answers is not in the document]
Gudahtt: Looks fantastic

guides/unit-testing.md
Gudahtt: We have quite a few tests in core that use fake timers selectively (e.g. the TokenRatesController). 

But regarding your question, a `withFakeTiimers` wrapper definitely makes sense. But the pattern I've used in the past is to have global teardown but not setup. e.g. call `jest.useRealTimers()` in `afterEach` but leave it up to each test to decide whether to enable fake timers or not. Similar to our typical `sinon.restore()` strategy.

I guess anytime we are "resetting" the global state back to its initial state, using `afterEach` seems fair game and unlikely to cause maintenance issues.

Gudahtt: Maybe the whole section should be updated to suggest avoiding `beforeEach`, and not mention `afterEach`? I'm having trouble thinking of a scenario where `afterEach` has caused problems.

guides/unit-testing.md
mcmire: Okay, I've updated this section. Hopefully these changes make more sense: 6646934...9f89251

guides/unit-testing.md
mcmire: Good point. Updated in 66469342083e36e0bc7b9e0bbecf9b15b65de259...9f892512f284e55f56defb2bdd136f675d546710.

guides/unit-testing.md
```diff
@@ -0,0 +1,973 @@
...
+```
+
+✅
+
+``` typescript
+it("starts the block tracker", () => {
+  // ...
+});
+
+it("returns a promise that resolves to a block number", () => {
```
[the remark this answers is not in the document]
mcmire: Ah, right. I guess this doesn't bother me that much 😅 but it makes sense. Updated in [2e1a567](https://github.com/MetaMask/contributor-docs/pull/38/commits/2e1a5673b99dbb3b4aafeec9a033d8d63bac0b41).

guides/unit-testing.md
mcmire: I ended up removing this section.

guides/unit-testing.md
mcmire: Okay fair! Updated in [7a34dfe](https://github.com/MetaMask/contributor-docs/pull/38/commits/7a34dfe959d7e2cfa57861cd660a42e61fc42665).

guides/unit-testing.md
mcmire: I've revised this section to talk primarily about `beforeEach`. I do include a section on handling `afterEach` if necessary but hid that (along with the full explanation) behind a disclosure. Updated here: [1a8381b](https://github.com/MetaMask/contributor-docs/pull/38/commits/1a8381b901963910688228d92853b647f988e5e5)

guides/unit-testing.md
mcmire: Extracted to a different section here: [1a8381b](https://github.com/MetaMask/contributor-docs/pull/38/commits/1a8381b901963910688228d92853b647f988e5e5)

guides/unit-testing.md
mcmire: Yes, I think I did mean to do that. I can update this.

mcmire: I also think that this example doesn't illustrate the problem that I'm referring to in the rationale. I actually think there are a couple reasons why it's good to keep critical data in the test. One is that as a reader you don't have to look so far to piece together the "story" the test is telling. The other (which I think you might be referring to here) is that it hides what's special about the data itself. That could be solved by better naming, though. Maybe there's another reason, though. I'll have to study real world code to refine this some more.

docs/sdlc.md
```diff
@@ -13,11 +13,12 @@ This lifecycle is intended to be used by software teams including product manage
 This lifecycle integrates security into all aspects of application software development which is comprised of six phases.
 
 1. Requirements Gathering and Planning:
-  - Identify the scope, goals, and necessary resources for development
-  - Define functional and security requirements for the feature 
-    - Considering [core engineering principles](./engineering-principles.md)
-    - Security controls such as authentication, data encryption, and secure communication protocols
-  - Inform the marketing team and customer support of the projected timelines, features, and any potential market impacts
+
```
mcmire, line 16: Don't worry about these changes — they will go away once #43 is merged.

docs/unit-testing.md
```diff
@@ -0,0 +1,1045 @@
...
+
+```
+src/
+  permission-controller.ts
+  permission-controller.test.ts
+```
+
+## Use `describe` to group tests for the module under test
+
+Wrap all tests for a file in a `describe` block named the same as the file itself:
```
Mrtenz, line 36: I'm curious why this is necessary? Jest will print the name of the file regardless.

Mrtenz (commented):
Another suggestion: Prefer `it.each` over a for-loop inside the test.

guides/unit-testing.md
Gudahtt: Yeah exactly. The password variable was completely sufficient for explaining both the "story" of the test and the conditions it was running under, because the test would apply to any arbitrary string as the password. If I had reason to double-check the value of "password" I may have been annoyed at having to scroll all the way up and back down again, but there was no need to look.

guides/unit-testing.md
Gudahtt: With that bullet, I was thinking of exactly the same ideas you shared in this comment: https://github.com/MetaMask/contributor-docs/pull/38#discussion_r1310946590

No need to cover that here if it's covered elsewhere. Those bullets were just examples to consider

docs/unit-testing.md
Gudahtt: Does it? I know it prints the name of the file where tests are currently running, but for the report I thought they all got combined

Gudahtt: Oh I see, no you're right, it always shows the filename in the console output. I'm not sure why we have this rule then.

I guess we need a `describe` block often enough for hooks, and it's nice to keep everything consistent. But maybe the strict requirement for having just one is a relic from when we used mocha.

guides/unit-testing.md
mcmire: I went through existing code in `core` and used that to update this example to something I feel like would be 1) more realistic and 2) genuinely annoying. Let me know what you think: [09116f7](https://github.com/MetaMask/contributor-docs/pull/38/commits/09116f79b63c1d0c057965ed451cfc45714edf8b)

Gudahtt: Oh my goodness, great example, that test is painful

guides/unit-testing.md
mcmire: Good point. I've updated this guideline here: [1decf7b](https://github.com/MetaMask/contributor-docs/pull/38/commits/1decf7b5e036099cd4a523f9f86b05e36fddd8ab)

docs/unit-testing.md
mcmire: > I guess we need a `describe` block often enough for hooks, and it's nice to keep everything consistent.

Right, that was exactly my thought. However, you're right, this isn't necessary, especially if we are already using `describe` to group tests for the same function/method. I can remove this.

mcmire: Done here: [34f61b6](https://github.com/MetaMask/contributor-docs/pull/38/commits/34f61b6f7913ac313b0772a3c21a34e1b7111e4e)

mcmire:
> Prefer `it.each` over a for-loop inside the test

@Mrtenz I'm going to be off for a while starting next week so I'm wondering if it makes sense to add this in a new PR? Plus this PR is pretty big 😅

mcmire:
I think I've addressed all the feedback above. I've made a few more tweaks along the way, so let me know if there's anything that doesn't make sense or could be simplified.

Gudahtt (approved):
LGTM!

Mrtenz (approved):

PR #39, opened 2023-08-17
Remove extraneous structure; add working TOC

mcmire:
This repository is fairly sparse at the moment, so while I understand that we will eventually want organization and consistency, I have two thoughts: 1) everyone may have a different idea on the kinds of documents that we want to write; and 2) everyone may have a different writing style. So I'm not sure that it makes sense to enforce any kind of structure yet. I think right now we ought to focus on making it as easy as possible for anyone (within the org, at least) to contribute.

To that end I've flattened the file structure so that all documents are now in a `docs/` directory. I've also removed the guidelines for guidelines, since I'm not sure that pattern will be used for all documents (and even as I've been adding more, I've noticed I'm not following my own rules, so I don't want to keep updating them until they're more set in stone).

I've also updated the README to be a bit friendlier and fixed the table of contents in the README, as they have been broken so far (now they just list all of the files in `docs/`). Also, I don't know that we need to necessarily say that this repository is in draft. It will always be living; I've highlighted that in the README as well.


README.md
```diff
@@ -1,12 +1,12 @@
...
\ No newline at end of file
+# MetaMask Contributor Documentation
+
+Welcome! Here you can find the principles, policies, and practices which are followed by MetaMask engineers in building MetaMask. These resources are designed to enable all builders, whether internal or external, in making meaningful contributions to any part of the MetaMask ecosystem.
+
+This is a living repository — nothing is set in stone! If there is a change you want to introduce, feel free to start a discussion in an issue or propose the change itself in a pull request.
+
+## Table of Contents
+
+- [Engineering Principles](./docs/engineering-principles.md)
```
Gudahtt, line 9: We'll want to add the SDLC document here as well

Gudahtt:
Sounds good. I like the structure but it does look silly at the moment. We can consider restoring it later once we have more content.

mcmire:
@sethkfman I noticed you made a change recently to remove DRAFT from all of the document headers. What do you think about this change?

Gudahtt (approved):
LGTM!

bdstar81 (approved):

bdstar81 (approved):

PR #41, opened 2023-08-21
Add lint step to dev and CI

mcmire:
- Add Yarn v3 (same as module template)
- Rely on Node 16 (same as module template)
- Add a private manifest, and add Prettier as a dev dependency
- Add the same `lint` script from the module template
- Use the same Prettier config that we use in the module template
- Run all Markdown files through Prettier, applying any auto-fixes
- Ensure `main` contains Markdown files that pass lint by configuring
  GitHub Actions to run the `lint` script


mcmire:
There are other projects we could use for linting Markdown:

- [`markdownlint`](https://github.com/DavidAnson/markdownlint)
- [`write-good`](https://github.com/btford/write-good)
- [`alex`](https://alexjs.com/)
- [`prose-lint`](http://proselint.com/)
- [`vale`](https://github.com/errata-ai/vale)

It'd also be cool if we could verify that all of the examples given are executable. There's a project called [`runmd`](https://github.com/broofa/runmd) that sort of does this, but it only works on JavaScript and is implemented in a fairly hacky way. I would be interested in potentially rewriting this using something like the [`unified`](https://unifiedjs.com/) toolkit to parse code blocks from Markdown, then transpile them via `swc` or something, then run them in a Node VM. But... I figured this was a good start for now!

Gudahtt:
Looks great! Lots of conflicts now though

mcmire:
Ready for review again.

Gudahtt (approved):
LGTM!

bdstar81 (approved):

bdstar81 (approved):

PR #58, opened 2023-10-24
docs: v1 secure coding guidelines

sethkfman:
This PR adds secure coding guidelines base on concepts from OWASP Top Ten and modified to the specific at MetaMask when developing client application and libraries. 

[Link to formatted version](https://github.com/MetaMask/contributor-docs/blob/chore/secure-coding-guidelines/docs/secure-coding-guidelines.md)

### DIFF
--- a/docs/secure-coding-guidelines.md
+++ b/docs/secure-coding-guidelines.md
@@ -57,8 +57,6 @@ The guidelines in this policy were gathered primarily from the [OWASP Top 10](ht
 
 #### Content Security
 
-- Verify the content type of external data to ensure it matches expectations
-- Enforce proper type checking, file type validation for file/media uploads and rendering third party content
 - Implement Content Security Policies (CSP) to mitigate XSS attacks when rendering external data
 
 #### Data Serialization
--- a/docs/secure-coding-guidelines.md
+++ b/docs/secure-coding-guidelines.md
@@ -10,7 +10,7 @@ These guidelines apply to developers building client based applications at MetaM
 
 ## References
 
-The guidelines in this policy were gathered from concepts on the [OWASP Top 10 Site](https://owasp.org/www-project-top-ten/) and modified to meet the client applications being developed at MetaMask.
+The guidelines in this policy were gathered primarily from the [OWASP Top 10](https://owasp.org/www-project-top-ten/) and the [OWASP Application Security Verification Standard](https://owasp.org/www-project-application-security-verification-standard/). We have included just the guidelines most relevant to MetaMask client applications.
 
 ## Guidelines
 
--- a/docs/secure-coding-guidelines.md
+++ b/docs/secure-coding-guidelines.md
@@ -30,7 +30,7 @@ The guidelines in this policy were gathered primarily from the [OWASP Top 10](ht
 #### 1.2 Content Security
 
 1. Verify the content type of external data to ensure it matches expectations
-2. Enforce proper type checking, file type validation for file/media uploads and rendering 3rd party content
+2. Enforce proper type checking, file type validation for file/media uploads and rendering third party content
 3. Implement Content Security Policies (CSP) to mitigate XSS attacks when rendering external data
 
 #### 1.3 Data Serialization
@@ -73,7 +73,7 @@ The guidelines in this policy were gathered primarily from the [OWASP Top 10](ht
 2. Errors should not leak data and be made more generic. This is especially important for code which handles data classified as **sensitive**.
 3. Production versions of applications shall have logging disabled
 
-### 4 3rd Party Integrations & Applications
+### 4 Third Party Integrations & Applications
 
 #### 4.1 Authentication and Authorization
 
--- a/docs/secure-coding-guidelines.md
+++ b/docs/secure-coding-guidelines.md
@@ -14,123 +14,123 @@ The guidelines in this policy were gathered primarily from the [OWASP Top 10](ht
 
 ## Guidelines
 
-### 1 Input from Users and Third Party Services
+### 1 – Input from Users and Third Party Services
 
-#### 1.1 Data Validation and Sanitization
+#### 1.1 – Data Validation and Sanitization
 
-1. Always validate and sanitize user inputs and reject any input that does not pass validation, such as:
+* 1.1.1: Always validate and sanitize user inputs and reject any input that does not pass validation, such as:
    1. Checking the type is what we expect: (e.g. is a javascript object being passed in when a string is expected)
    2. Checking if the value is what within a range we expect: (e.g. is a negative number being provided when a positive one is expected)
    3. Checking if the format is what we expect: (e.g. do we expect a `0x` prefixed hex string but an alphanumeric string is provided instead)
    4. Checking if a value matches an allowed set of constants (e.g. is the value one of the few possible numbers we expect)
-2. Validate and sanitize data from external sources before rendering it in the application's UI
-3. Preference should be given to libraries and frameworks that support input validation
-4. Avoid dynamic code execution with untrusted data to prevent injection attacks
+* 1.1.2: Validate and sanitize data from external sources before rendering it in the application's UI
+* 1.1.3: Preference should be given to libraries and frameworks that support input validation
+* 1.1.4: Avoid dynamic code execution with untrusted data to prevent injection attacks
 
-#### 1.2 Content Security
+#### 1.2 – Content Security
 
-1. Verify the content type of external data to ensure it matches expectations
-2. Enforce proper type checking, file type validation for file/media uploads and rendering third party content
-3. Implement Content Security Policies (CSP) to mitigate XSS attacks when rendering external data
+* 1.2.1: Verify the content type of external data to ensure it matches expectations
+* 1.2.2: Enforce proper type checking, file type validation for file/media uploads and rendering third party content
+* 1.2.3: Implement Content Security Policies (CSP) to mitigate XSS attacks when rendering external data
 
-#### 1.3 Data Serialization
+#### 1.3 – Data Serialization
 
-1. Recommend using pure data formats such as CSV, JSON for serialization to avoid deserialization attacks
+* 1.3.1: Recommend using pure data formats such as CSV, JSON for serialization to avoid deserialization attacks
 
-### 2 Data Persistence & Encryption
+### 2 – Data Persistence & Encryption
 
-#### 2.1 Data Classification
+#### 2.1 – Data Classification
 
-1. Classify the data in your application into at least two groups:
-2. **Non-sensitive** data (e.g. publicly available, transactions, app theme settings, etc…)
-3. **Sensitive** data (e.g. passwords, keys)
+* 2.1.1: Classify the data in your application into at least two groups:
+* 2.1.2: **Non-sensitive** data (e.g. publicly available, transactions, app theme settings, etc…)
+* 2.1.3: **Sensitive** data (e.g. passwords, keys)
 
-#### 2.2 Data Security
+#### 2.2 – Data Security
 
-1. Encrypt **sensitive** data prior to persisting
-2. Limit the amount of time **sensitive** data is decrypted
-3. Do not bundle API tokens in client side applications unless you intend for them to be publicly accessible
+* 2.2.1: Encrypt **sensitive** data prior to persisting
+* 2.2.2: Limit the amount of time **sensitive** data is decrypted
+* 2.2.3: Do not bundle API tokens in client side applications unless you intend for them to be publicly accessible
 
-#### 2.3 Encryption
+#### 2.3 – Encryption
 
-1. **Sensitive** data should be secured at rest and in-transit using standard encryption protocols (e.g. cryptography libraries, OS level key management systems, https, etc…)
-2. Cryptography of **sensitive** data should meet a minimum standard established by the security team during a threat assessment and annual reviews
+* 2.3.1: **Sensitive** data should be secured at rest and in-transit using standard encryption protocols (e.g. cryptography libraries, OS level key management systems, https, etc…)
+* 2.3.2: Cryptography of **sensitive** data should meet a minimum standard established by the security team during a threat assessment and annual reviews
 
-#### 2.4 Passwords
+#### 2.4 – Passwords
 
-1. Passwords used to decrypt content should adhere to security team recommendations during a threat assessment and annual reviews.
+* 2.4.1: Passwords used to decrypt content should adhere to security team recommendations during a threat assessment and annual reviews.
 
-### 3 Logging & Error Handling
+### 3 – Logging & Error Handling
 
-#### 3.1 Logging
+#### 3.1 – Logging
 
-1. Secrets and **sensitive** data should not be logged
-2. Console logging should be removed from production code using linters
+* 3.1.1: Secrets and **sensitive** data should not be logged
+* 3.1.2: Console logging should be removed from production code using linters
 
-#### 3.2 Error Handling
+#### 3.2 – Error Handling
 
-1. All errors should be handled
-2. Errors should not leak data and be made more generic. This is especially important for code which handles data classified as **sensitive**.
-3. Production versions of applications shall have logging disabled
+* 3.2.1: All errors should be handled
+* 3.2.2: Errors should not leak data and be made more generic. This is especially important for code which handles data classified as **sensitive**.
+* 3.2.3: Production versions of applications shall have logging disabled
 
-### 4 Third Party Integrations & Applications
+### 4 – Third Party Integrations & Applications
 
-#### 4.1 Authentication and Authorization
+#### 4.1 – Authentication and Authorization
 
-1. Implement access controls and authentication mechanisms for accessing external data
-2. Store and manage API keys securely when accessing external sources
-3. API keys used in client applications shall have a reduced scope adhering to Principle of Least Privilege
-4. If an API key is not intended to be publicly accessible, do not use it in a client application
+* 4.1.1: Implement access controls and authentication mechanisms for accessing external data
+* 4.1.2: Store and manage API keys securely when accessing external sources
+* 4.1.3: API keys used in client applications shall have a reduced scope adhering to Principle of Least Privilege
+* 4.1.4: If an API key is not intended to be publicly accessible, do not use it in a client application
 
-#### 4.2 Integration/Application Integrity
+#### 4.2 – Integration/Application Integrity
 
-1. Monitor integrations/applications for changes and security issues.
+* 4.2.1: Monitor integrations/applications for changes and security issues.
 
-#### 4.3 Permissions
+#### 4.3 – Permissions
 
-1. All applications/dapps/snaps shall adhere to the following to Principle of Least Privilege with a default state of allowing no permissions
-2. Users shall provide consent for required permissions of applications/dapps/snaps
+* 4.3.1: All applications/dapps/snaps shall adhere to the following to Principle of Least Privilege with a default state of allowing no permissions
+* 4.3.2: Users shall provide consent for required permissions of applications/dapps/snaps
 
-### 5 Dependency Management
+### 5 – Dependency Management
 
-#### 5.1 Dependency Integrity
+#### 5.1 – Dependency Integrity
 
-1. Use Dependabot to keep all dependencies up to date and to automate checks for updates
-2. Pin dependency versions or use a yarn.lock file
+* 5.1.1: Use Dependabot to keep all dependencies up to date and to automate checks for updates
+* 5.1.2: Pin dependency versions or use a yarn.lock file
 
-#### 5.2 Avoid Deprecated and Unmaintained Packages
+#### 5.2 – Avoid Deprecated and Unmaintained Packages
 
-1. Using Socket.dev to check the health and maintenance status of dependencies and seek alternatives when necessary
+* 5.2.1: Using Socket.dev to check the health and maintenance status of dependencies and seek alternatives when necessary
 
-#### 5.3 Review Dependency Licenses
+#### 5.3 – Review Dependency Licenses
 
-1. Ensure dependencies' licenses align with your project's requirements
+* 5.3.1: Ensure dependencies' licenses align with your project's requirements
 
-#### 5.4 Reduce Dependencies and Keep Them Minimal
+#### 5.4 – Reduce Dependencies and Keep Them Minimal
 
-1. Minimize dependencies to reduce potential vulnerabilities
+* 5.4.1: Minimize dependencies to reduce potential vulnerabilities
 
-#### 5.5 Audit and Monitor Dependencies
+#### 5.5 – Audit and Monitor Dependencies
 
-1. Projects should use the npm audit, Dependabot and Socket.dev tooling for periodic automated auditing of dependencies
+* 5.5.1: Projects should use the npm audit, Dependabot and Socket.dev tooling for periodic automated auditing of dependencies
 
-#### 5.6 LavaMoat
+#### 5.6 – LavaMoat
 
-1. LavaMoat allow-scripts should be enabled on all projects
-2. LavaMoat runtime should be enabled on all projects
+* 5.6.1: LavaMoat allow-scripts should be enabled on all projects
+* 5.6.2: LavaMoat runtime should be enabled on all projects
 
-### 6 Operations & Infrastructure
+### 6 – Operations & Infrastructure
 
-#### 6.1 Project secrets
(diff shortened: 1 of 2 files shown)


docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,96 @@
...
+The guidelines in this policy were gathered from concepts on the [OWASP Top 10 Site](https://owasp.org/www-project-top-ten/) and modified to meet the client applications being developed at MetaMask.
+
+## Guidelines
+
+This document identifies six categories developers should reference when working on client based applications.
+
+1. User Input:
+
+   1. Processing User Inputs:
+      1. Always validate and sanitize user inputs
```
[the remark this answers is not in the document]
sethkfman: Updated and added your examples

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,96 @@
...
+      2. Preference should be given to libraries and frameworks that support input validation
```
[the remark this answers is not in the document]
sethkfman: This may be something that should be moved down to dependencies. The general idea here is that is we are selecting libraries that process user data there should be a preference for the library/framework that performs validation over one that doesn't.

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,96 @@
...
+   2. Data Serialization:
+      1. Recommend using pure data formats such as JSON for serialization to avoid deserialization attacks
```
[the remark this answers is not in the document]
sethkfman: Adde CSV

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,96 @@
...
+
+2. Data Persistence & Encryption:
+
+   1. Data Classification:
+      1. Classify the data in your application into at least two groups:
+      1. Non-sensitive data (e.g. publicly available, transactions, tokens, etc…)
```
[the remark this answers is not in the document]
sethkfman: Ah, good call out. Replace with `app theme settings`

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,96 @@
...
+      1. Sensitive data (e.g. passwords, keys)
```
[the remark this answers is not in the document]
sethkfman: Corrected

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,96 @@
...
+      1. Passwords used to decrypt content should adhere to security team recommendations during a threat assessment and annual reviews.
+
+3. Logging & Error Handling:
+
+   1. Logging:
+      1. Secrets and sensitive data should not be logged
+      2. Console logging should be removed from production code using linters
+   2. Error Handling:
+      1. All errors should be handled
+      2. Errors should not leak data and be made more generic
```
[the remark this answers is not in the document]
sethkfman: Added

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,96 @@
...
+
+4. 3rd Party Data & Applications:
+
+   1. Data Validation and Sanitization:
+      1. Validate and sanitize data from external sources before rendering it in the application's UI
+      2. Avoid dynamic code execution with untrusted data to prevent injection attacks
+   2. Content Security:
+      1. Verify the content type of external data to ensure it matches expectations
+      2. Enforce proper type checking, file type validation for file/media uploads and rendering 3rd party content
+      3. Implement Content Security Policies (CSP) to mitigate XSS attacks when rendering external data
```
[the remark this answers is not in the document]
sethkfman: I reorganized things a bit. Let me know what you think.

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,99 @@
...
+      1. Use Dependabot to keep all dependencies up to date and to automate checks for updates
+      2. Pin dependency versions or use a yarn.lock file
+   2. Avoid Deprecated and Unmaintained Packages:
+      1. Using Socket.dev to check the health and maintenance status of dependencies and seek alternatives when necessary
+   3. Review Dependency Licenses:
+      1. Ensure dependencies' licenses align with your project's requirements
+   4. Reduce Dependencies and Keep Them Minimal:
+      1. Minimize dependencies to reduce potential vulnerabilities
+   5. Audit and Monitor Dependencies:
+      1. Projects should use the npm audit, Dependabot and Socket.dev tooling for periodic automated auditing of dependencies
```
desi, line 86: I am unfamiliar with Socket.dev. Is this something we have configured already on our repos? And if so is "periodic automated auditing" already defined for most repos and if so is there a "safe default" for what the frequency should be? I am still learning how everything is setup and just wondering how a new dev reading this would know if this is already being handled or if its something they need to do.

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,99 @@
...
+
+5. Dependency Management:
+
+   1. Dependency Integrity:
+      1. Use Dependabot to keep all dependencies up to date and to automate checks for updates
+      2. Pin dependency versions or use a yarn.lock file
+   2. Avoid Deprecated and Unmaintained Packages:
+      1. Using Socket.dev to check the health and maintenance status of dependencies and seek alternatives when necessary
+   3. Review Dependency Licenses:
+      1. Ensure dependencies' licenses align with your project's requirements
```
desi, line 82: Is there a tool that helps with this that we actively recommend or is it expected this is a manual check?

desi (commented):
I left a few questions and a small typo call out. It looks great though and nothing I wrote is a blocker.

docs/secure-coding-guidelines.md
sethkfman: @desi Great question. Socket.dev is a new tool that is run on all MetaMask org PRs. Here is a bit more information: https://www.notion.so/Socket-dev-usage-recommendations-379c24adff6a40a5987b854d4a174691

docs/secure-coding-guidelines.md
sethkfman: There are tools out there. @Gudahtt Had recommended automating this aspect but at this time it is a manual check.

docs/secure-coding-guidelines.md
naugtur, line 86:
```suggestion
      1. Projects should use tools (npm audit, Dependabot and Socket.dev) for periodic automated auditing of dependencies
      2. Use npm audit locally as a tool that informs you of the backlog of issues you have. Use dependabot security alerts to react to new vulnerability reports in your dependencies.
        - Triage dependabot PRs early, merge quickly if updates are minor. Don't hesitate to close them if not useful or could never be applied - but leave a comment why. Call attention to PRs where a security update is a major version update and cannot be easily introduced.
      3. If vulnerability management is particularly important for the project, run audit as a CI step and fail if it finds new issues. This may be prone to false positives and be blocked by transitive dependencies not being updated upstream. Use `resolutions` to override or `npm-audit-resolver` package to build a process around safely ignoring selected issues (requires discipline)
      4. Socket.dev will automatically report on suspicious dependencies in your pull requests. For best results use the following etiquette when dealing with it:
        - avoid merging the PR without addressing Socket warnings
        - avoid the ignore-all command unless you're introducing socket to a new project that's known to be safe (very rare)
        - always add a short comment explaining why you're ignoring a warning of potential malice/danger. eg. for a publisher change 'known maintainer' is enough of a comment indicating that you've verified the new publisher of the package is legitimate.
        - if Socket is being annoying, remember i can be configured. contact @naugtur (Zbyszek) and report if a warning category is useless.
```

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,99 @@
...
+   4. Reduce Dependencies and Keep Them Minimal:
+      1. Minimize dependencies to reduce potential vulnerabilities
+   5. Audit and Monitor Dependencies:
+      1. Projects should use the npm audit, Dependabot and Socket.dev tooling for periodic automated auditing of dependencies
+   6. LavaMoat:
+      1. LavaMoat allow-scripts should be enabled on all projects
+      2. LavaMoat runtime should be enabled on all projects
```
naugtur, line 89:
```suggestion
      2. LavaMoat runtime should be enabled on all projects
        - runtime protections for node.js running your build
        - LavaMoat tools for bundling
      3. When you regenerate the policy file look carefully at the diff. Raise an issue with your team and don't hesitate to call for help in the #lavamoat channel if new items in the policy look like a lot of new privileges or include the ability to **spawn/exec child processes**
      4. Review your policy.json for suspicious permissions at least once in your participation in a project (eg. when you join a new project, familiarize yourself with it and then look at the policy. Ask your team questions as you go.)
```

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,99 @@
...
+
+6. Operations & Infrastructure:
+   1. Project secrets:
+      1. Project secrets should be generated and shared using a password management system (1Password) and access to that system should follow the Principle of Least Authority
+      2. Build systems needing access to project secrets should store secrets in specific secret storage and obscured once placed into system
```
naugtur, line 94:
```suggestion
      2. Build systems needing access to project secrets should store secrets in specific secret storage and obscured once placed into system
      3. If your setup allows that, use a tool like `2ms` (too many secrets) to detect whether you're attempting to commit a secret to the repository.
```

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,99 @@
...
+   1. Dependency Integrity:
+      1. Use Dependabot to keep all dependencies up to date and to automate checks for updates
+      2. Pin dependency versions or use a yarn.lock file
+   2. Avoid Deprecated and Unmaintained Packages:
+      1. Using Socket.dev to check the health and maintenance status of dependencies and seek alternatives when necessary
+   3. Review Dependency Licenses:
+      1. Ensure dependencies' licenses align with your project's requirements
+   4. Reduce Dependencies and Keep Them Minimal:
+      1. Minimize dependencies to reduce potential vulnerabilities
+   5. Audit and Monitor Dependencies:
```
naugtur, line 85:
```suggestion
   5. Audit and Monitor Dependencies:
     0. Discuss an establish a process for maintaining your dependencies and dealing with old dependencies you get stuck on because of breaking changes.
```

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,99 @@
...
+1. Input from Users and Third Party Services:
+   1. Data Validation and Sanitization:
+      1. Always validate and sanitize user inputs and reject any input that does not pass validation, such as:
+         1. Checking the type is what we expect: (e.g. is a javascript object being passed in when a string is expected)
+         2. Checking if the value is what within a range we expect: (e.g. is a negative number being provided when a positive one is expected)
+         3. Checking if the format is what we expect: (e.g. do we expect a `0x` prefixed hex string but an alphanumeric string is provided instead)
+         4. Checking if a value matches an allowed set of constants (e.g. is the value one of the few possible numbers we expect)
+      2. Validate and sanitize data from external sources before rendering it in the application's UI
+      3. Preference should be given to libraries and frameworks that support input validation
+      4. Avoid dynamic code execution with untrusted data to prevent injection attacks
```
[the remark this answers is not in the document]
Gudahtt: Good point. I've expanded on this a bit here: https://github.com/MetaMask/contributor-docs/pull/58/commits/98bdca213e2aeea477d7f86fa78ae99f1123e8a1
Let me know what you think!

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,99 @@
...
+This document identifies six categories developers should reference when working on client based applications.
+
+1. Input from Users and Third Party Services:
+   1. Data Validation and Sanitization:
+      1. Always validate and sanitize user inputs and reject any input that does not pass validation, such as:
+         1. Checking the type is what we expect: (e.g. is a javascript object being passed in when a string is expected)
+         2. Checking if the value is what within a range we expect: (e.g. is a negative number being provided when a positive one is expected)
+         3. Checking if the format is what we expect: (e.g. do we expect a `0x` prefixed hex string but an alphanumeric string is provided instead)
+         4. Checking if a value matches an allowed set of constants (e.g. is the value one of the few possible numbers we expect)
+      2. Validate and sanitize data from external sources before rendering it in the application's UI
```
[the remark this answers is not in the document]
Gudahtt: Good idea! Added here: https://github.com/MetaMask/contributor-docs/pull/58/commits/b86b880a9bd7ec66aedb1010de2d0412e21333d8  and here: https://github.com/MetaMask/contributor-docs/pull/58/commits/9ac6a746fa1f9b5139467120571920818a9e705d

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,99 @@
...
+
+This document identifies six categories developers should reference when working on client based applications.
+
+1. Input from Users and Third Party Services:
+   1. Data Validation and Sanitization:
+      1. Always validate and sanitize user inputs and reject any input that does not pass validation, such as:
+         1. Checking the type is what we expect: (e.g. is a javascript object being passed in when a string is expected)
+         2. Checking if the value is what within a range we expect: (e.g. is a negative number being provided when a positive one is expected)
+         3. Checking if the format is what we expect: (e.g. do we expect a `0x` prefixed hex string but an alphanumeric string is provided instead)
+         4. Checking if a value matches an allowed set of constants (e.g. is the value one of the few possible numbers we expect)
```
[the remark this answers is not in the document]
Gudahtt: I'm not sure I understand this one. Even for snaps, I thought all communication was via `postMessage`, so the data should be serialized (i.e. there'd be no way to define any getters as part of the data). Maybe I'm missing something.

docs/secure-coding-guidelines.md
Gudahtt: Thanks, great suggestions! I haven't taken this exactly as-is, but I've tried to capture the spirit of it here: 43e5282ad7cc2c6c1a78c66b68f8b95286b25056

It's been condensed a bit, and the Dependabot part is omitted. I omitted Dependabot because I tend not to think of it as a security tool; it's quite slow and unreliable, sometimes taking many days to create a PR to address vulnerabilities, and it will silently fail to address them sometimes. It's also often not configured or not configured correctly in our organization. I left a "Update dependencies quickly" bullet there to nudge people to follow your advice about responding quickly though.

docs/secure-coding-guidelines.md
Gudahtt: Thanks! I've added something like this here: 62164d7

I'd like to follow up to add some examples of suspicious permissions though. I didn't include the spawn/exec example yet. I want to confirm exactly how they'd show up in the policy first.

docs/secure-coding-guidelines.md
Gudahtt: Good idea, I've added something like this here: 0162d1808e7a54bee352f46c4643a44d1192458c

docs/secure-coding-guidelines.md
Gudahtt: Hmm, this is a curious one. I'm not aware of any teams that have an explicit process like this, so I'm hesitant to suggest we all should have one. Would like to discuss this further though.

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,172 @@
...
+  - Periodically scan for security vulnerabilities (e.g. using tools like `npm audit`)
+  - Update dependencies quickly when they have security vulnerabilities
+  - Use Socket.dev to monitor dependencies for other noteworthy changes, such as maintainer changes, or the addition of install scripts or binary files
+    - Use the following etiquette when addressing Socket.dev warnings:
+      - Investigate and address all warnings before merging a PR
+      - Avoid using the `ignore-all` bot command, instead ignoring each warning one at a time
+      - If you've investigated a warning and found that it's not indicative of a malicious dependency, ignore it with a bot comment and explain your investigation with a short comment
+      - Contact the security team if you're unsure how to investigate something, or if you'd like to disable a warning category
+
+#### LavaMoat
```
sethkfman, line 143: What do you think about calling out LavaMoat specific to JS?

```suggestion
#### LavaMoat (JavaScript projects only)
```

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,172 @@
...
+- Store and manage project secrets securely
+  - Each CI system should have a method of securely managing secrets
+  - For secrets used directly by developers, use 1Password
+- Limit access to project secrets, following the Principle of Least Authority
+- Prevent contributors from committing secrets to the repository
+  - GitHub has a "Secret scanning" setting that can help with this, and there are third-party tools like `2ms` (too many secrets) that can help as well
+
+#### Application Integrity
+
+- Application should be distributed through approved store portals (GitHub, App Store, Play Store, Firefox Store, Chrome Store)
```
sethkfman, line 171:
```suggestion
- Application should be distributed through approved store portals (e.g. GitHub, App Store, Play Store, Firefox Store, Chrome Store)
```

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,172 @@
...
+- Each distribution channel shall have provide release verifiability (certificate signing approval, hash verification)
```
sethkfman, line 172:
```suggestion
- Each distribution channel shall have provide release verifiability (e.g. certificate signing approval, hash verification)
```

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,172 @@
...
+
+  - HTML-encode strings before they are included in the DOM
+    - Some libraries do this automatically (e.g. React)
+  - URL-encode data before including it in a URL
+
+- When accepting a URI as input, ensure the scheme matches expectations
+
+  For example, a URL for a website or API would typically have a scheme of `https`
+
+- Avoid dynamic code execution with untrusted data
```
sethkfman, line 46: Do you think we should add a statement? 

If dynamic code execution is a required part of a feature, a threat model and Security Team review should occur.

docs/secure-coding-guidelines.md
Gudahtt: Sure, makes sense

docs/secure-coding-guidelines.md
Gudahtt: Makes sense, I'll try to work this in

Gudahtt: Done here: 3f7883e584889119ca21c85d7afbc071b7015d7d

Gudahtt (approved):
LGTM!

docs/secure-coding-guidelines.md
naugtur: the dependabot part was based on actual problems we've recently had. dependabot is not a tool helping to react fast. It's a reminder and it does some of the work for you creating the PR. It's a tool helping not to stay many versions behind on stuff. And we've had dependabot PRs for security updates sit there for over a year.

docs/secure-coding-guidelines.md
naugtur: This was the most important point I wanted to make here. Insert a mandate to start the conversations across teams. I will push for this via various means, but I was hoping to get it in front of everyone and this first edition of the training was a great opportunity.

docs/secure-coding-guidelines.md
```diff
@@ -0,0 +1,172 @@
...
+    - Use the following etiquette when addressing Socket.dev warnings:
+      - Investigate and address all warnings before merging a PR
+      - Avoid using the `ignore-all` bot command, instead ignoring each warning one at a time
+      - If you've investigated a warning and found that it's not indicative of a malicious dependency, ignore it with a bot comment and explain your investigation with a short comment
+      - Contact the security team if you're unsure how to investigate something, or if you'd like to disable a warning category
+
+#### LavaMoat (JavaScript projects only)
+
+- LavaMoat `allow-scripts` should be enabled on all projects
+  - This project relies upon install scripts being disabled in your package manager. This can be verified by adding the dependency `@lavamoat/preinstall-always-fail`, which will cause installation to fail if scripts are enabled.
```
naugtur, line 146: `allow-scripts setup` adds that dependency. Also, yarn 3 and 4 will not fail the entire install process because they swallow errors.

docs/secure-coding-guidelines.md
naugtur:
```
always add a short comment explaining why you're ignoring a warning of potential malice/danger. eg. for a publisher change 'known maintainer' is enough of a comment indicating that you've verified the new publisher of the package is legitimate.
```
became
```
If you've investigated a warning and found that it's not indicative of a malicious dependency, ignore it with a bot comment and explain your investigation with a short comment
```
The latter is IMHO too vague to be memorable and I'm afraid people won't internalize what it means and either ignore the recommendation or put too much effort into creating writeups leading to annoyance.

This is my main issue with this doc - it's too terse and open to interpretation and could use much more examples.

docs/secure-coding-guidelines.md
Gudahtt: Ah I see, I hadn't realized that about Yarn v3/v4. I guess your point here is that it'd be safer to advise using `allow-scripts setup` because users might fail to properly disable install scripts in the package manager otherwise, is that correct? Just wanted to confirm that setting `ignore-scripts` is also something handled by the setup script.

docs/secure-coding-guidelines.md
Gudahtt: Part of the reason that I'm having trouble with this is that I don't fully understand the suggestion. Specifically it's the condition you think teams should have a plan for, this "stuck on old dependencies" situation. Most often the only obstacle to updating dependencies is a little investment in time, so it's rare that we enter a situation where we're "stuck".

Did you have any examples in mind of a situation where we'd benefit from having a process like this, and what that process might look like?

docs/secure-coding-guidelines.md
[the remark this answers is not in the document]
Gudahtt: The  "did not pass serialization before" part is what I'm not following. Is that maybe the wrong description of this scenario? Or is there a case where messages aren't serialized that I'm not aware of.

docs/secure-coding-guidelines.md
Gudahtt: Point taken on the wording, I'll restore something closer to yours. Can't recall the thinking behind that right now.

For the most part the terseness of the document isn't intended, we just haven't had time to expand on most points yet. More detail would definitely be appreciated.

Gudahtt: I appreciate that dependabot has been serving an important role absent other reminders about security problems, but I think there are good reasons to prefer alternatives. We haven't rolled out an npm audit CI job yet in libraries, but it's a far more reliable solution. Dependabot on the other hand is extremely slow, frequently fails to update things correctly, and will silently give up if it doesn't find a way to update something. It's not a reliable way to flag security advisories. I intended to add an npm audit CI job to the module template to serve that purpose.

docs/secure-coding-guidelines.md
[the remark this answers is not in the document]
naugtur: messages coming from `ppom` and from a snap are being validated before  they're passed to the actual RPC that JSON.stringifies them. as a result validation can be cheated with a getter. Also, JSON serialization that happens within RPC provider must not trust the object it serializes.

example of a recent regression in snaps: https://osec.io/blog/2023-11-01-metamask-snaps/#safe-json-exploit

docs/secure-coding-guidelines.md
naugtur: Dependabot is only useful together with the alerts, yes

docs/secure-coding-guidelines.md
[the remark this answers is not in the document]
Gudahtt: Ah of course, our code must be running in the same process as the Compartment in order to construct it correctly, and messages at that layer have never been serialized. Of course, not sure how I missed that. Thanks for the link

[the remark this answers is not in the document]
Gudahtt: This suggestion has been restored here: https://github.com/MetaMask/contributor-docs/pull/84

docs/secure-coding-guidelines.md
Gudahtt: The example Socket.dev resolution was restored here, along with a second example: https://github.com/MetaMask/contributor-docs/pull/85

I started restoring the Dependabot advice in this PR but didn't get far: https://github.com/MetaMask/contributor-docs/pull/86

It felt wrong to include advice on triaging Dependabot issues given that it's not used in many repos at the moment, and even in those where it is enabled it's often not setup correctly (such that the Dependabot PRs typically don't pass CI checks).

PR #61, opened 2023-11-06
fix: Secure Coding Path

sethkfman:
The Secure Coding Guidelines document path in the table of contents is broken. The path has been updated and is valid.


sethkfman:
FIx in another PR

PR #62, opened 2023-11-06
fix: the secure coding doc path

sethkfman:
The Secure Coding Guidelines document path in the table of contents is broken. The path has been updated and is valid.


Gudahtt (approved):
LGTM!

PR #64, opened 2023-11-09
SDLC: Fix formatting of lists

mcmire:
Promote six phases into sections so that the lists within these phases sit at the right level.


Gudahtt (approved):
LGTM!

bdstar81 (approved):

PR #68, opened 2023-12-21
Add Shared Libraries team in codeowners file

desi:
Based on Github Teams Proposal we want to update codeowners to reflect the team(s) that should be consulted before a file is changed. In this case we are adding the Shared Libraries team.


MajorLift (dismissed):

mcmire (approved):

.github/CODEOWNERS
```diff
@@ -1,4 +1,4 @@
 # Lines starting with '#' are comments.
 # Each line is a file pattern followed by one or more owners.
 
-*                                    @MetaMask/engineering
+*                                   @MetaMask/shared-libraries-engineers
```
Mrtenz, line 4:
```suggestion
*                                   @MetaMask/shared-libraries-engineers

```

desi: Just pushed fix

Issue #69, opened 2024-01-03
Complete TypeScript guidelines


mcmire:
We've started a document which is intended to contain TypeScript guidelines, but it is severely lacking in content.

We should get to an MVP by adding the following guidelines:

- Use enums instead of union types
- Avoid type assertions, or explain their usage when necessary

mcmire:
Off the top of my head, here's a list of guidelines we could include:

- Don't use numeric enums
  - Numeric enums are misleading because it creates a reverse mapping from value to property name, making `Object.values` not do what you think it does
- Constrain generic types if necessary
  - It may not be enough just to have a type or a function take another type — you might have to constrain it if it's not allowed to be anything (e.g. `extends Json`)
- 💡 Use `Partial` to type the accumulating variable in `reduce`
- 💡 Use `Omit` to reduce requirements
  - If you want to reuse a type for, say, an argument to a function, but you only use part of that type, then you can use `Omit` to reduce that type to a subset. This makes it easier to test the function because you don't have to create a full version of the object that represents the original type.
- Use enums instead of union types
  - Inevitably you will want to refer to the values of a union type somewhere (perhaps as the argument to a function). You can of course just use a literal which represents a member of that union — but if you have an enum, then all of the values are special, and any time you use a value then anyone can see where that value comes from.
- Don't use `any`
  - `any` doesn't actually represent "any type"; if you declare a value as `any`, it tells TypeScript to disable typechecking for that value anywhere it appears.
  - `any` is dangerous because it infects all surrounding and downstream code. For instance, if you have a function that is declared to return `any` which actually returns an object, all properties of that object will be `any`.
  - Need to truly represent something as "any value"? Use `unknown` instead.
- Avoid type assertions, or explain why they are needed
  - Notable exceptions: `reduce`, iterating over the keys of an enum
- Avoid type annotations
  - Notable exceptions: `reduce` (?)
  - Need to create a data structure with specific values, but ensure that it's declared with a more abstract type? Create a function that takes an argument of the type you want, then use it to build the object. (Note: This will be solved in a future version of TypeScript with the `satisfies` keyword.)

There are also a couple of tickets we've created already:

- https://github.com/MetaMask/contributor-docs/issues/47
- https://github.com/MetaMask/contributor-docs/issues/57

desi:
To get to an MVP we are going to do the enum bullet, don't use any bullet and the type assertions bullet. We will come back to additional items later. Pointing based on this comment.

[...]

PR #74, opened 2024-02-07
TypeScript Guidelines - Types: Inference, Annotations, Assertions, Escape Hatches

### LINKED ISSUE #47: Add entry for`any` usage in TypeScript Guidelines doc
## Motivation
- There are currently hundreds of instances of `any` type usage and `as any` assertions spread throughout [metamask-extension](https://github.com/MetaMask/metamask-extension) and [core](https://github.com/MetaMask/core).
- For future maintainability, there should be clear guidelines defining the limited cases where `any` usage is intentional and acceptable vs. the majority of cases where `any` should be corrected and removed. 

## Outline
- This is a non-exhaustive list of cases where `any` usage might be acceptable:
  - In tests, to type mock objects or intentionally break features.
  - To pass in as a generic type argument. (more detail?)

### LINKED ISSUE #57: Add TS guideline: "Document safe or necessary uses of type assertions"
When a type assertion is absolutely necessary due to constraints or is even safe due to runtime checks, we should document the reason for doing so. Something like:

``` typescript
// Type assertion: We ensure that "xyzzy" is filled in above.
const foo = bar as Baz;
```

MajorLift:
## Motivation

In order to improve type safety and maintainability, we need to establish clear guidelines regarding how to apply types: specifically, when and when not to use explicit type declarations or keywords such as `as`, `any`.

## Explanation

- See markdown preview: [TypeScript Guidelines - "Types" section](https://github.com/MetaMask/contributor-docs/blob/240206-typescript-guidelines-any-as/docs/typescript.md)
- ~The merge conflicts with main will be resolved after the review process to avoid noisy diffs.~
- All examples are tagged with a permalink for easy reference and quotation.

### Table of Contents
- Types
  - Type Inference
  - Type Annotations (`:`, `satisfies`)
  - Type Assertions (`as`, `!`)
  - Escape Hatches (`any`, `@ts-expect-error`)
    
### Details
- Dangers and advantages.
- Tips for avoiding.
- Acceptable use cases.

## References
- Closes #47
- Closes #57 
- Contributes to #69

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -563,7 +563,21 @@ exampleFunction('__proto__');
 
 #### If accompanied by a TODO comment, `@ts-expect-error` is acceptable to use for marking errors that have clear plans of being resolved
 
-<!-- TODO: Add example -->
+**Example <a id="example-43313247-4393-4966-b78e-378f636fedec"></a> ([🔗 permalink](#example-43313247-4393-4966-b78e-378f636fedec)):**
+
+✅
+
+```typescript
+// @ts-expect-error TODO: remove this annotation once the `Eip1193Provider` class is released, resolving thi provider misalignment issue.
+return new Web3Provider(provider);
+
+// TODO: Fix this by handling or eliminating the undefined case
+// @ts-expect-error This variable can be `undefined`, which would break here.
+```
+
+This recommendation applies to any disruptive change that creates many errors at once (e.g. dependency update, upstream refactor, package migration).
+
+See [this entry](https://github.com/MetaMask/core/blob/main/docs/package-migration-process-guide.md#4-resolve-or-todo-downstream-errors) in the core repo "package migration process guide," which recommends that complex or blocked errors should be annotated with a `// @ts-expect-error TODO:` comment, and then revisited once the disruptive change has been completed.
 
 #### Always avoid `any`
 
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -15,9 +15,9 @@ Note that this document assumes that the reader has a high level of familiarity 
 TypeScript provides a range of syntax for communicating type information with the compiler.
 
 - The compiler performs **type inference** on all types and values in the code.
-- The user can assign **type annotations** (`:`, `satisfies`) to override inferred types or add type constraints.
-- The user can add **type assertions** (`as`, `!`) to force the compiler to accept user-supplied types even if they contradicts the inferred types.
-- Finally, there are **escape hatches** that let type checking be disabled (`any`, `@ts-expect-error`) for a certain scope of code.
+- The user can assign **type annotations** (`satisfies`, `:`) to override inferred types or add type constraints.
+- The user can add **type assertions** (`!`, `as`) to force the compiler to accept user-supplied types even if they contradicts the inferred types.
+- Finally, there are **escape hatches** that let type checking be disabled (`@ts-expect-error`, `any`) for a certain scope of code.
 
 The order of this list represents the general order of preference for using these features.
 
@@ -509,7 +509,7 @@ This is often the case when downstream consumers of the code are using JavaScrip
 
 🚫
 
-> **Error:** This comparison appears to be unintentional because the types '`0x${string}`' and '"\_\_proto\_\_"' have no overlap.ts(2367)
+> **Error:** This comparison appears to be unintentional because the types '\`0x${string}\`' and '"\_\_proto\_\_"' have no overlap.ts(2367)
 
 ```typescript
 function exampleFunction(chainId: `0x${string}`) {
@@ -522,7 +522,7 @@ function exampleFunction(chainId: `0x${string}`) {
 
 🚫
 
-> **Error:** Argument of type '"\_\_proto\_\_"' is not assignable to parameter of type '`0x${string}`'.ts(2345)
+> **Error:** Argument of type '"\_\_proto\_\_"' is not assignable to parameter of type '\`0x${string}\`'.ts(2345)
 
 ```typescript
 exampleFunction('__proto__');
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -25,7 +25,7 @@ There are several reasons for this:
 
 🚫 Type declarations
 
-```ts
+```typescript
 const name: string = 'METAMASK'; // Type 'string'
 
 const BUILT_IN_NETWORKS = new Map<string, `0x${string}`>([
@@ -36,7 +36,7 @@ const BUILT_IN_NETWORKS = new Map<string, `0x${string}`>([
 
 ✅ Type inferences
 
-```ts
+```typescript
 const name = 'METAMASK'; // Type 'METAMASK'
 
 const BUILT_IN_NETWORKS = {
@@ -47,7 +47,7 @@ const BUILT_IN_NETWORKS = {
 
 #### Example: To annotate or not to annotate
 
-```ts
+```typescript
 type TransactionMeta = TransactionBase &
   (
     | {
@@ -73,7 +73,7 @@ this.messagingSystem.publish(
 
 🚫 Add type annotation
 
-```ts
+```typescript
 // Type 'TransactionMeta'
 const updatedTransactionMeta: TransactionMeta = {
   ...transactionMeta,
@@ -83,7 +83,7 @@ const updatedTransactionMeta: TransactionMeta = {
 
 ✅ Add `as const` and leave to inference
 
-```ts
+```typescript
 // Type narrower than 'TransactionMeta': { status: TransactionStatus.rejected; ... }
 // (doesn't include 'error' property)
 const updatedTransactionMeta = {
@@ -104,7 +104,7 @@ There is a clear exception to the above: if an explicit type annotation or asser
 
 🚫
 
-```ts
+```typescript
 const chainId: string = this.messagingSystem(
   'NetworkController:getProviderConfig',
 ).chainId; // Type 'string'
@@ -112,7 +112,7 @@ const chainId: string = this.messagingSystem(
 
 ✅
 
-```ts
+```typescript
 const chainId = this.messagingSystem(
   'NetworkController:getProviderConfig',
 ).chainId; // Type '`0x${string}`'
@@ -124,14 +124,14 @@ This is one case where type inference is unable to reach a useful conclusion wit
 
 🚫
 
-```ts
+```typescript
 const tokens = []; // Type 'any[]'
 const tokensMap = new Map(); // Type 'Map<any, any>'
 ```
 
 ✅
 
-```ts
+```typescript
 const tokens: string[] = []; // Type 'string[]'
 const tokensMap = new Map<string, Token>(); // Type 'Map<string, Token>'
 ```
@@ -140,7 +140,7 @@ const tokensMap = new Map<string, Token>(); // Type 'Map<string, Token>'
 
 <!-- TODO: Add explanation and examples -->
 
-```ts
+```typescript
 function isSomeInterface(x: unknown): x is SomeInterface {
   return (
     'name' in x &&
@@ -211,7 +211,7 @@ Otherwise, only mocking the properties needed in the test improves readability b
 
 <!-- TODO: Add examples -->
 
-```ts
+```typescript
 const handler:
   | ((payload_0: ComposableControllerState, payload_1: Patch[]) => void)
   | ((payload_0: any, payload_1: Patch[]) => void); // Type of 'payload_0': 'any'
@@ -236,7 +236,7 @@ Some generic types use `any` as a default generic argument. This can silently in
 
 🚫
 
-```ts
+```typescript
 const NETWORKS = new Map({
   mainnet: '0x1',
   goerli: '0x5',
@@ -251,7 +251,7 @@ mockGetNetworkConfigurationByNetworkClientId.mockImplementation(
 
 ✅
 
-```ts
+```typescript
 const NETWORKS = new Map<string, `0x${string}`>({
   mainnet: '0x1',
   goerli: '0x5',
@@ -276,7 +276,7 @@ In most type errors involving property access or runtime property assignment, `a
 
 🚫
 
-```ts
+```typescript
 for (const key of getKnownPropertyNames(this.internalConfig)) {
   (this as any)[key] = this.internalConfig[key];
 }
@@ -288,7 +288,7 @@ delete addressBook[chainId as any];
 
 ✅
 
-```ts
+```typescript
 for (const key of getKnownPropertyNames(this.config)) {
   (this as unknown as typeof this.config)[key] = this.config[key];
 }
@@ -300,20 +300,20 @@ However, when assigning to a generic type, using `as any` is the only solution.
 
 🚫
 
-```ts
+```typescript
 (state as RateLimitState<RateLimitedApis>).requests[api][origin] = previous + 1;
 // is generic and can only be indexed for reading.ts(2862)
 ```
 
 ✅
 
-```ts
+```typescript
 (state as any).requests[api][origin] = previous + 1;
 ```
 
 Even in this case, however, `any` usage might be avoidable by using `Object.assign` or spread operator syntax instead of assignment.
 
-```ts
+```typescript
 Object.assign(state, {
   requests: {
     ...state.requests,
@@ -334,7 +334,7 @@ Object.assign(state, {
 
 ✅
 
-```ts
+```typescript
 class BaseController<
   ...,
   messenger extends RestrictedControllerMessenger<N, any, any, string, string>
@@ -362,7 +362,7 @@ Although TypeScript is capable of inferring return types, adding them explicitly
 
 🚫
 
-```ts
+```typescript
 async function removeAccount(address: Hex) {
   const keyring = await this.getKeyringForAccount(address);
 
@@ -379,7 +379,7 @@ async function removeAccount(address: Hex) {
 
 ✅
 
-```ts
+```typescript
 async function removeAccount(address: Hex): Promise<KeyringControllerState> {
   const keyring = await this.getKeyringForAccount(address);
 
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -2,13 +2,320 @@
 
 These guidelines specifically apply to TypeScript.
 
-## Provide explicit return types for functions/methods
+## Types
+
+### Type Inference
+
+TypeScript is very good at inferring types. A well-maintained codebase can provide strict type safety with explicit type annotations being the exception rather than the rule.
+
+When writing TypeScript, function and class signatures, and types that express the domain model of the codebase will require explicit definitions and annotations.
+
+However, for other types and values that are derived from those fundamentals, type inferences should generally be preferred over type declarations and assertions.
+
+There are several reasons for this:
+
+#### Advantages
+
+- Explicit type annotations (`:`) and type assertions (`as`, `!`) prevent inference-based narrowing of the user-supplied types.
+  - The compiler errs on the side of trusting user input, which prevents it from providing additional, and sometimes crucial, type information that it could otherwise infer.
+  <!-- TODO: Expand into entry. Add example. -->
+  - In TypeScript v4.9+ the `satisfies` operator can be used to assign type constraints that are narrowable through type inference.
+- Type inferences are responsive to changes in code without requiring user input, while annotations and assertions rely on hard-coding, making them brittle against code drift.
+- The `as const` operator can be used to further narrow an inferred abstract type into a specific literal type.
+
+🚫 Type declarations
+
+```ts
+const name: string = 'METAMASK'; // Type 'string'
+
+const BUILT_IN_NETWORKS = new Map<string, `0x${string}`>([
+  ['mainnet', '0x1'],
+  ['goerli', '0x5'],
+]); // Type 'Map<string, `0x${string}`>'
+```
+
+✅ Type inferences
+
+```ts
+const name = 'METAMASK'; // Type 'METAMASK'
+
+const BUILT_IN_NETWORKS = {
+  mainnet: '0x1',
+  goerli: '0x5',
+} as const; // Type { readonly mainnet: '0x1'; readonly goerli: '0x5'; }
+```
+
+### Type Narrowing
+
+There is a clear exception to the above: if an explicit type annotation or assertion can narrow an inferred type further, thereby improving its accuracy, it should be applied.
+
+##### Avoid widening a type with a type annotation
+
+> **Warning**<br />
+> Double-check that a declared type is narrower than the inferred type.<br />
+> Enforcing an even wider type defeats the purpose of adding an explicit type annotation, as it _loses_ type information instead of adding it.
+
+🚫
+
+```ts
+const chainId: string = this.messagingSystem(
+  'NetworkController:getProviderConfig',
+).chainId; // Type 'string'
+```
+
+✅
+
+```ts
+const chainId = this.messagingSystem(
+  'NetworkController:getProviderConfig',
+).chainId; // Type '`0x${string}`'
+```
+
+##### When instantiating an empty container type, provide a type annotation
+
+This is one case where type inference is unable to reach a useful conclusion without user-provided information. Since the compiler cannot arbitrarily restrict the range of types that could be inserted into the container, it has to assume the widest type, which is often `any`. It's up to the user to narrow that into the intended type with an explicit type annotation.
+
+🚫
+
+```ts
+const tokens = []; // Type 'any[]'
+const tokensMap = new Map(); // Type 'Map<any, any>'
+```
+
+✅
+
+```ts
+const tokens: string[] = []; // Type 'string[]'
+const tokensMap = new Map<string, Token>(); // Type 'Map<string, Token>'
+```
+
+##### Type guards and null checks can be used to improve type inference
+
+<!-- TODO: Add explanation and examples -->
+
+```ts
+function isSomeInterface(x: unknown): x is SomeInterface {
+  return (
+    'name' in x &&
+    typeof x.name === 'string' &&
+    'length' in x &&
+    typeof x.length === 'number'
+  );
+}
+
+function f(x: SomeInterface | SomeOtherInterface) {
+  if (isSomeInterface(x)) {
+    // Type of x: 'SomeInterface | SomeOtherInterface'
+    console.log(x.name); // Type of x: 'SomeInterface'. Type of x.name: 'string'.
+  }
+}`
+```
+
+### Type Assertions
+
+`as` assertions are unsafe. They overwrite type-checked and inferred types with user-supplied types that suppress compiler errors.
+
+#### Avoid `as`
+
+Type assertions make the code brittle against changes. While TypeScript will throw type errors against some unsafe or structurally unsound type assertions, it will generally accept the user-supplied type without type-checking. This can cause silent failures where errors are suppressed, even though the type's relationship to the rest of the code, or the type itself, has been altered so that the type assertion is no longer valid.
+
+#### Acceptable usages of `as`
+
+##### To prevent or fix `any` usage
+
+Unsafe as type assertions may be, they are still categorically preferable to using `any`.
+
+- With type assertions, we still get working intellisense, autocomplete, and other IDE features.
+- Type assertions also provide an indication of the expected type as intended by the author.
+- For type assertions to an incompatible shape, use `as unknown as` as a last resort rather than `any` or `as any`.
+
+##### To define user-defined type guards
+
+`as` syntax is often required to write type guards. See https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates.
+
+##### To type data objects whose shape and contents are determined at runtime
+
+Preferably, this typing should be accompanied by schema validation performed with type guards and unit tests.
+
+- e.g. The output of `JSON.parse()` or `await response.json()` for a known JSON input.
+- e.g. The type of a JSON file.
+
+<!-- TODO: Add example -->
+
+##### In tests, for mocking or to exclude irrelevant but required properties from an input object
+
+<!-- TODO: Add examples -->
+
+It's recommended to provide accurate typing if there's any chance that omitting properties affects the accuracy of the test.
+
+Otherwise, only mocking the properties needed in the test improves readability by making the intention and scope of the mocking clear, not to mention being convenient to write.
+
+### Compiler Directives
+
+<!-- TODO: Add section for `@ts-expect-error` -->
+
+#### Avoid `any`
+
+`any` is the most dangerous form of explicit type declaration, and should be completely avoided if possible.
+
+- `any` doesn't represent the widest type, or indeed any type at all. `any` is a compiler directive for _disabling_ static type checking for the value or type to which it's assigned.
+- `any` suppresses all error messages about its assignee. This includes errors that are changed or newly introduced by alterations to the code. This makes `any` the cause of dangerous **silent failures**, where the code fails at runtime but the compiler does not provide any prior warning.
+- `any` subsumes all other types it comes into contact with. Any type that is in a union, intersection, is a property of, or has any other relationship with an `any` type or value is erased and becomes an `any` type itself.
+
+<!-- TODO: Add examples -->
+
+```ts
+const handler:
+  | ((payload_0: ComposableControllerState, payload_1: Patch[]) => void)
+  | ((payload_0: any, payload_1: Patch[]) => void); // Type of 'payload_0': 'any'
+```
+
+- `any` pollutes all surrounding and downstream code.
+<!-- TODO: Add examples -->
+
+#### Fixes for `any`
+
+##### Try `unknown` instead
+
+`unknown` is the universal supertype i.e. the widest possible type.
+
+- When typing the assignee, `any` and `unknown` are interchangeable (every type is assignable to both).
+- However, when typing the assigned, `unknown` can't be used to replace `any`, as `unknown` is only assignable to `unknown`. In this case, try `never`, which is assignable to all types.
+<!-- TODO: Add example -->
+
+##### Don't allow generic types to use `any` as a default argument
+
+Some generic types use `any` as a default generic argument. This can silently introduce an `any` type into the code, causing unexpected behavior and suppressing useful errors.
+
+🚫
+
+```ts
+const NETWORKS = new Map({
+  mainnet: '0x1',
+  goerli: '0x5',
+}); // Type 'Map<any, any>'
+
+const mockGetNetworkConfigurationByNetworkClientId = jest.fn(); // Type 'jest.Mock<any, any>'
+mockGetNetworkConfigurationByNetworkClientId.mockImplementation(
+  (origin, type) => {},
+); // No error!
+// Even though 'mockImplementation' should only accept callbacks with a signature of '(networkClientId: string) => NetworkConfiguration | undefined'
+```
+
+✅
+
+```ts
+const NETWORKS = new Map<string, `0x${string}`>({
+  mainnet: '0x1',
+  goerli: '0x5',
+}); // Type 'Map<string, `0x${string}`>'
+
+const mockGetNetworkConfigurationByNetworkClientId = jest.fn<
+  ReturnType<NetworkController['getNetworkConfigurationByNetworkClientId']>,
+  Parameters<NetworkController['getNetworkConfigurationByNetworkClientId']>
+>(); // Type 'jest.Mock<NetworkConfiguration | undefined, [networkClientId: string]>'
+mockGetNetworkConfigurationByNetworkClientId.mockImplementation(
+  (origin, type) => {},
+);
+// Argument of type '(origin: any, type: any) => void' is not assignable to parameter of type '(networkClientId: string) => NetworkConfiguration | undefined'.
+// Target signature provides too few arguments. Expected 2 or more, but got 1.ts(2345)
+```
+
+#### Acceptable usages of `any`
+
+##### Assigning new properties to a generic type at runtime
+
+In most type errors involving property access or runtime property assignment, `any` usage can be avoided by substituting with `as unknown as`.
+
+🚫
+
+```ts
+for (const key of getKnownPropertyNames(this.internalConfig)) {
+  (this as any)[key] = this.internalConfig[key];
+}
+
+delete addressBook[chainId as any];
+// Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ [chainId: `0x${string}`]: { [address: string]: AddressBookEntry; }; }'.
+//  No index signature with a parameter of type 'string' was found on type '{ [chainId: `0x${string}`]: { [address: string]: AddressBookEntry; }; }'.ts(7053)
+```
+
+✅
+
+```ts
+for (const key of getKnownPropertyNames(this.config)) {
+  (this as unknown as typeof this.config)[key] = this.config[key];
+}
+
+delete addressBook[chainId as unknown as `0x${string}`];
+```
+
+However, when assigning to a generic type, using `as any` is the only solution.
+
+🚫
+
+```ts
+(state as RateLimitState<RateLimitedApis>).requests[api][origin] = previous + 1;
+// is generic and can only be indexed for reading.ts(2862)
+```
+
+✅
+
+```ts
+(state as any).requests[api][origin] = previous + 1;
+```
+
+Even in this case, however, `any` usage might be avoidable by using `Object.assign` or spread operator syntax instead of assignment.
+
+```ts
+Object.assign(state, {
+  requests: {
+    ...state.requests,
+    [api]: { [origin] },
+  },
+});
+
+{
+  ...state,
+  requests: {
+    ...state.requests,
+    [api]: { [origin] },
+  },
+};
+```
+
+##### Within generic constraints
+
+✅
+
+```ts
+class BaseController<
+  ...,
+  messenger extends RestrictedControllerMessenger<N, any, any, string, string>
+> ...
+```
+
+- In general, using `any` in this context is not harmful in the same way that it is in other contexts, as the `any` types only are not directly assigned to any specific variable, and only function as constraints.
+- That said, more specific constraints provide better type safety and intellisense, and should be preferred wherever possible.
+
+##### Catching errors
+
+- `catch` only accepts `any` and `unknown` as the error type.
+- Recommended: Use `unknown` with type guards like `isJsonRpcError`.
+- Avoid typing an error object with `any` if it is passed on to be used elsewhere instead of just being thrown, as the `any` type will infect the downstream code.
+
+##### In tests, for mocking or to intentionally break features
+
+<!-- TODO: Add examples -->
+
+## Functions
+
+##### For functions and methods, provide explicit return types
 
 Although TypeScript is capable of inferring return types, adding them explicitly makes it much easier for the reader to see the API from the code alone and prevents unexpected changes to the API from emerging.
 
 🚫
 
-```javascript
+```ts
 async function removeAccount(address: Hex) {
   const keyring = await this.getKeyringForAccount(address);
 
@@ -25,7 +332,7 @@ async function removeAccount(address: Hex) {
 
 ✅
 
-```javascript
+```ts
 async function removeAccount(address: Hex): Promise<KeyringControllerState> {
   const keyring = await this.getKeyringForAccount(address);
 
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -669,15 +669,22 @@ To prevent `any` instances from being introduced into the codebase, it is not en
   🚫 A single type, `InferWithParams`, is set to `any` in `@metamask/utils`
 
   ```typescript
-  export declare type InferWithParams<Type extends Struct<any>, Params extends JsonRpcParams> = any;
-  
-  export declare type JsonRpcRequest<Params extends JsonRpcParams = JsonRpcParams> = InferWithParams<typeof JsonRpcRequestStruct, Params>; // Resolves to 'any'
-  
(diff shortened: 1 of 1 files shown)


docs/typescript.md
```diff
@@ -2,13 +2,504 @@
...
+- `never` is the universal subtype i.e. the narrowest possible type, equivalent to the null set(∅).
+- `never` is assignable to every type, but the only type that is assignable to `never` is `never`.
+- When typing the _assigned_:
+  - `unknown` is unable to replace `any`, as `unknown` is only assignable to `unknown`.
+  - The type of the _assigned_ must be a subtype of the _assignee_.
+  - `never` is worth trying, as it is the universal subtype and assignable to all types.
+
+<!-- TODO: Add examples -->
+
+##### Don't allow `any` to be used as a generic default argument
```
NicolasMassart, line 369: Q: wouldn't it be enough to just keep the paragraph explaining any must not be used because of its disabling of type checks and not add more? There's too many text about do not use any for this or than when just do not use any first explanation is probably enough.

docs/typescript.md
```diff
@@ -2,13 +2,504 @@
...
+
+Type assertions make the code brittle against changes.
+
+While TypeScript and ESLint will flag some unsafe, structurally unsound, or redundant type assertions, they will generally accept user-supplied types without further type-checking.
+
+This can cause silent failures or false negatives where errors are suppressed. This is especially damaging as the codebase accumulates changes over time. Type assertions may continue to silence errors, even though the type itself, or the type's relationship to the rest of the code may have been altered so that the asserted type is no longer valid.
+
+##### Redundant or unnecessary `as` assertions are not flagged for removal
+
+Type assertions can also cause false positives, because they assertions are independent expressions, untied to the type errors they were intended to fix. Thus, even if code drift fixes or removes a particular type error, the type assertions that were put in place to fix that error will provide no indication that they are no longer necessary and now should be removed.
```
NicolasMassart, line 171:
```suggestion
Type assertions can also cause false positives, because assertions are independent expressions, untied to the type errors they were intended to fix. Thus, even if code drift fixes or removes a particular type error, the type assertions that were put in place to fix that error will provide no indication that they are no longer necessary and now should be removed.
```

docs/typescript.md
```diff
@@ -2,13 +2,504 @@
...
+  )
+  .forEach((elm) =>
+    this.updateNft(
+      elm.value.nft,
+      ...
+    ),
+  );
+```
+
+#### Acceptable usages of `as`
```
NicolasMassart, line 263: C: we often have to use as to type a process.env value that is always considered a string. Maybe this could be added as example?

NicolasMassart (commented):
Great explanation. A bit too verbose in my opinion, and contains too complicated sentences (think about non native English readers).
I feel that the part on any is too long and goes too deep in details when we should just never use it.

docs/typescript.md
MajorLift: Absolutely! This sounds like a great example for an externally determined type. Would you be able to provide an example for this?

docs/typescript.md
```diff
@@ -2,13 +2,504 @@
...
+##### Try `unknown` and `never` instead
+
+###### `unknown`
+
+- `unknown` is the universal supertype i.e. the widest possible type, equivalent to the universal set(U).
+- Every type is assignable to `unknown`, but `unknown` is only assignable to `unknown`.
+- When typing the _assignee_, `any` and `unknown` are completely interchangeable since every type is assignable to both.
+- `any` usage is often motivated by a need to find a placeholder type that could be anything. `unknown` is the most likely type-safe substitute for `any` in these cases.
+
+###### `never`
```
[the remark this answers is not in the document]
MajorLift: Any suggestions on what changes might improve readability here? 

I agree it's a tricky topic, but the purpose of the guidelines isn't to instruct readers about TypeScript, but to convey preferred practices to readers who are already knowledgeable on that subject. So I'm a little fuzzy on how much detail or ELI5 analogies I should be providing.

Also, is "C:" short for change? I assume "Q:" is for question?

docs/typescript.md
MajorLift: We do have a eslint rule for banning `any` usage now, and [usually](https://github.com/MetaMask/contributor-docs/pull/80#discussion_r1507074598) I would agree that conventions that we enforce with linters or static code analysis do not require detailed explanations or justifications.

However, I strongly feel that `any` is a special case that warrants special attention.
- The epics and issues attached to this PR should show that writing guidelines for `any` usage is the main scope of this ticket.
- The tickets attached to those epics will show that a significant portion of the work shared libraries does is fixing existing `any` usage and preventing introduction of new instances through code review.
- This is all because `any` is such a convenient escape hatch that there is a strong incentive to override the eslint ban and use it anyway. Simply reiterating the ban is not enough to be effective.
- Therefore, the guidelines must be written to cultivate a deeper understanding in our contributors about exactly why and how `any` usage is destructive to a codebase.

docs/typescript.md
MajorLift: Thanks for pointing out the awkward phrasing here.

```suggestion
Type assertions can also cause false positives, because they are independent expressions, untied to the type errors they are intended to fix. Thus, even if code drift fixes or removes a particular type error, the type assertions that were put in place to fix that error will provide no indication that they are no longer necessary and should now be removed.
```

docs/typescript.md
MajorLift: I added a short explanation summarizing the above: https://github.com/MetaMask/contributor-docs/pull/74/files#diff-ae03a9f2eb638a25c1d8d7ba853c78220f1ed44a19ad97a8d780cbc0c00a4260R411-R413

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+### Type Annotations
+
+An explicit type annotation is acceptable for overriding an inferred type if...
+
+1. It can further narrow an inferred type, thus supplying type information that the compiler cannot infer or access.
+2. It is being used to enforce a type constraint rather than assign a type definition.
+
+##### Use the `satisfies` operator to enforce a type constraint or to validate the assigned type
+
+In the second case, an abstract, "narrowest supertype" is used to constrain or validate a type. TypeScript provides the `satisfies` operator for this use case. It is able to both enforce a type constraint and further narrow the assigned type through inference.
```
Gudahtt, line 129: Nit: The "In the second case" here caught me off guard the first time, I thought it was talking about the second example rather than the second case in the above section. Maybe this could be simplified:
```suggestion
TypeScript provides the `satisfies` operator for constraining or validating a type. It is able to both enforce a type constraint and further narrow the assigned type through inference.
```

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+  ...transactionMeta,
+  status: TransactionStatus.rejected as const,
+} satisfies TransactionMeta;
+
+updatedTransactionMeta.error; // Property 'error' does not exist on type '{ status: TransactionStatus.rejected; ... }'.(2339)
+```
+
+#### Acceptable usages of `:` annotations
+
+##### Use type annotations if it will prevent or remove type assertions
```
Gudahtt, line 155: Hmmm.... not sure we need this guideline. I agree that annotations are preferable to assertions, but it's confusing to say that this is an acceptable use of annotations without clearer context for what the assertion had been used for in the first place.

I'm getting the implication that you had in mind a "type constraint/validation" scenario, but now that we're starting to use newer versions of TypeScript with the `satisfies` operator, we don't need annotations for those either.

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+```typescript
+const updatedTransactionMeta = {
+  ...transactionMeta,
+  status: TransactionStatus.rejected as const,
+} satisfies TransactionMeta;
+
+updatedTransactionMeta.error; // Property 'error' does not exist on type '{ status: TransactionStatus.rejected; ... }'.(2339)
+```
+
+#### Acceptable usages of `:` annotations
```
Gudahtt, line 153: Nit: I know that type signatures are referenced as an acceptable use earlier in the document, but it might be worth briefly mentioning this again here as well, so we don't mistakenly create the opposite impression

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+
+`as` assertions are inherently unsafe. They overwrite type-checked and inferred types with user-supplied types that suppress compiler errors.
+
+They should only be introduced into the code if the accurate type is unreachable through other means.
+
+##### Document safe or necessary use of type assertions
+
+When a type assertion is absolutely necessary due to constraints or is even safe due to runtime checks, we should document the reasoning behind its usage in the form of a comment.
+
+<!-- TODO: Add example -->
```
Gudahtt, line 191: Maybe we could use the Immer draft type assertion here, where we're using the assertion as a workaround to a bug.

https://github.com/MetaMask/core/blob/7ccd8e9f1ef63cefbada3d96fdac5c7bd836fcb0/packages/base-controller/src/BaseControllerV2.ts#L195

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+```typescript
+enum Direction {
+  Up = 'up',
+  Down = 'down',
+  Left = 'left',
+  Right = 'right',
+}
+const directions = Object.values(Direction);
+
+// Error: Element implicitly has an 'any' type because index expression is not of type 'number'.(7015)
```
Gudahtt, line 214: I'm having trouble following this example. What is this error meant to represent - is this the type error that we should see here?

When I try this in the TypeScript Playground, I get this:

```
Conversion of type 'string[]' to type 'keyof directions[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.(2352)
```

Gudahtt: Oh I see, this error was present but was eliminated by the type assertions. Got it. Maybe we could add some context here to explain that.

Also, I'll leave a separate comment for the error I saw, I may have a solution

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+// Only one of the two `as` assertions necessary to fix error, but neither are flagged as redundant.
+for (const key of Object.keys(directions) as keyof directions[]) {
```
Gudahtt, line 216:
```suggestion
for (const key of Object.keys(directions) as (keyof typeof directions)[]) {
```

This fixes the TS2352 error I saw on this line in the playground

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+    typeof x.length === 'number'
+  );
+}
+```
+
+🚫 Type assertion
+
+```typescript
+function f(x: SomeInterface | SomeOtherInterface) {
+  console.log((x as SomeInterface).name);
```
Gudahtt, line 238: Nit: This might make for a more realsitic example of what not to do:
```suggestion
  if (x.name) {
    // We know that x is `SomeInterface` because `x` has a `name` but `SomeOtherInterface` does not
    console.log((x as SomeInterface).name);
  }

```

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+}
+```
+
+✅ Narrowing with type guard
+
+```typescript
+function f(x: SomeInterface | SomeOtherInterface) {
+  if (isSomeInterface(x)) {
+    // Type of x: 'SomeInterface | SomeOtherInterface'
```
Gudahtt, line 247: Shouldn't this comment be outside of the `if` block?

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+    console.log(x.name); // Type of x: 'SomeInterface'. Type of x.name: 'string'.
+  }
+}
+```
+
+```typescript
```
Gudahtt, line 253: This is another place where the examples run together

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+// Error: Argument of type '"getNftInformation"' is not assignable to parameter of type 'keyof NftController'.ts(2345)
+// 'getNftInformation' is a private method of class 'NftController'
+sinon.stub(nftController, 'getNftInformation');
+```
+
+🚫 `as any`
+
+```typescript
+sinon.stub(nftController, 'getNftInformation' as any);
```
Gudahtt, line 316: It's unclear to me why we'd use `as any` here in the first place.

Maybe we can remove this guideline as well? This is another case where the guideline is about dealing with bad TypeScript, rather than legitimate scenarios where `as` is needed

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+- Even an assertion to a wrong type still allows the compiler to show us warnings and errors as the code changes, and is therefore preferrable to the dangerous radio silence enforced by `any`.
+- For type assertions to an incompatible shape, use `as unknown as` as a last resort rather than `any` or `as any`.
+
+##### `as` is always acceptable to use for TypeScript syntax other than type assertion
+
+- Writing type guards often requires using the `as` keyword.
+
+```typescript
+function isFish(pet: Fish | Bird): pet is Fish {
+  return (pet as Fish).swim !== undefined;
```
Gudahtt, line 334: I don't think this is an example of "TypeScript syntax other than type assertion". This is type assertion. It's also avoidable even in this example (using our `hasProperty` helper)

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+✅ Compiler specifies that the target type should be `keyof NftController`
+
+```typescript
+sinon.stub(nftController, 'getNftInformation' as keyof typeof nftController);
+```
+
+- Even an assertion to a wrong type still allows the compiler to show us warnings and errors as the code changes, and is therefore preferrable to the dangerous radio silence enforced by `any`.
+- For type assertions to an incompatible shape, use `as unknown as` as a last resort rather than `any` or `as any`.
+
+##### `as` is always acceptable to use for TypeScript syntax other than type assertion
```
Gudahtt, line 328: Hmmm. This is a good callout. I found this unexpected though, because I was reading the "Acceptable usages of `as`" header for this section as "Acceptable uses of `as` type assertion", since we're in the type assertion section. This isn't an example of acceptable type assertion, it's not type assertion at all.

Maybe we could move this up to the `####` level, in a section just clarifying that this isn't at type assertion and the guidelines here do not apply.

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+
+```typescript
+type MappedTypeWithNewProperties<Type> = {
+  [Properties in keyof Type as NewKeyType]: Type[Properties];
+};
+```
+
+##### `as` is acceptable to use for typing data objects whose shape and contents are determined at runtime, externally, or through deserialization
+
+Preferably, this typing should be accompanied by runtime schema validation performed with type guards and unit tests.
```
Gudahtt, line 348: If we're doing runtime validation, we should get the types from this validation as well.


I don't really see the case for using type assertions here unless we're skipping validation intentionally for performance reasons.

Gudahtt: Even if we're knowingly skipping validation for performance reasons, really this would only be safe to do if it was an internal API where we know exactly how it's being used. For example, our inter-process communication in the extension.

The guideline says "determined at runtime, externally, or through deserialization", but it's just the deserialization scenario I was able to think of. Using type assertions for things determined at runtime or externally seems really unsafe to me.

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+
+- e.g. The output of `JSON.parse()` or `await response.json()` for a known JSON input.
+- e.g. The type of a JSON file.
+
+<!-- TODO: Add example -->
+
+##### `as` may be acceptable to use in tests, for mocking or to exclude irrelevant but required properties from an input object
+
+It's recommended to provide accurate typing if there's any chance that omitting properties affects the accuracy of the test.
```
Gudahtt, line 357: Nit: This is a strange thing to say. Of course providing an invalid type would affect the accuracy of the test, how could it not? And how are we expecting people to make this determination?


When it comes to using `as` in tests, I had expected the guideline to be something like "`as` may be used to intentionally pass in invalid types in tests"

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+
+TypeScript provides several directive comments that can be used to suppress TypeScript compiler errors. Using these to ignore typing issues is dangerous and reduces the overall effectiveness of TypeScript.
+
+#### Acceptable usages of `@ts-expect-error`
+
+##### Use `@ts-expect-error` to force runtime execution of a branch for validation or testing
+
+Sometimes, there is a need to force a branch to execute at runtime for security or testing purposes, when that branch has correctly been inferred as being inaccessible by the TypeScript compiler.
+
+Suppressing compiler errors to avoid typing issues is usually dangerous because it results in a loss of information and safety. However, consciously using `@ts-expect-error` to add runtime checks represents an improvement in these aspects, and therefore is not discouraged.
```
Gudahtt, line 373: Nit: In general it's fair to say that suppressing compiler errors "results in a loss of information and safety", but I don't think that applies to unreachable branches. I don't think that's adequately explained here ("prepresents an improvement in these aspects" is pretty vague), but I can't think of a great way to explain it concisely either, and it doesn't seem totally necessary. So maybe we can drop this paragraph instead:

```suggestion
```

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+// @ts-expect-error Suppressing to test runtime error handling
+
+// @ts-expect-error Intentionally testing invalid state
+
+// @ts-expect-error We are intentionally passing bad input.
+```
+
+##### If accompanied by a TODO comment, `@ts-expect-error` is acceptable to use for marking errors that have clear plans of being resolved
+
+<!-- TODO: Add example -->
```
Gudahtt, line 403: Hmm. For this I was thinking of the cases where a TypeScript update results in new type errors, we could introduce a `@ts-expect-error` as a temporary fix so that we can resolve the type errors in a separate PR from the TypeScript update.

Maybe we can use the v4.4 update as the example here. It included changes to the `catch(error)` type that had a widespread impact, as you can see here: https://github.com/MetaMask/core/pull/795

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+- Every type is assignable to `unknown`, but `unknown` is only assignable to `unknown`.
+- When typing the _assignee_, `any` and `unknown` are completely interchangeable since every type is assignable to both.
+
+###### `never`
+
+`never` is worth trying as a starting point for narrowing down the type, as it is the universal subtype and assignable to all types.
+
+- `never` is the universal subtype i.e. the narrowest possible type, equivalent to the null set(∅).
+- `never` is assignable to every type, but the only type that is assignable to `never` is `never`.
+- When typing the _assigned_:
```
Gudahtt, line 452: What does "the assigned" mean here? I'm having trouble understanding this collection of points

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+);
+// Argument of type '(origin: any, type: any) => void' is not assignable to parameter of type '(networkClientId: string) => NetworkConfiguration | undefined'.
+// Target signature provides too few arguments. Expected 2 or more, but got 1.ts(2345)
+```
+
+#### Acceptable usages of `any`
+
+`any` is only acceptable for very specific usages. The following should be considered an exhaustive list of its valid use cases.
+
+##### `any` may be necessary when assigning new properties to or deleting properties from a generic type at runtime
```
Gudahtt, line 501: Hmm, I don't think I've encountered this situation before. Has this come up? I'm not sure why we'd want to do this.

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+class BaseController<
+  ...,
+  messenger extends RestrictedControllerMessenger<N, any, any, string, string>
+> ...
+```
+
+- In general, using `any` in this context is not harmful in the same way that it is in other contexts, as the `any` types only are not directly assigned to any specific variable, and only function as constraints.
+- That said, more specific constraints provide better type safety and intellisense, and should be preferred wherever possible.
+
+##### `any` may be acceptable to use to type the error property in a catch block
```
Gudahtt, line 576: Thoughts on removing this guideline? Using `unknown` seems easy enough to me that I see no reason to use `any` instead, given that we have utility functions to help with narrowing errors to make use of error properties like `message`.

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+
+- `catch` only accepts `any` and `unknown` as the error type.
+- Recommended: Use `unknown` with type guards like `isJsonRpcError`.
+- Avoid typing an error object with `any` if it is passed on to be used elsewhere instead of just being thrown, as the `any` type will infect the downstream code.
+
+##### `any` may be acceptable to use in tests, to intentionally break features
+
+<!-- TODO: Add examples. Should be clear why type assertions couldn't be used instead -->
```
Gudahtt, line 584: Generally I don't see any major issue with using `any` in tests, but I am curious when it would be preferred over assertions or `@ts-expect-error`. I've been trying to use `@ts-expect-error` for any intentionally-invalid type cases myself.

docs/typescript.md
MajorLift: I wrote a short explanation about assignee and assigned with examples here: https://github.com/MetaMask/contributor-docs/issues/69#issuecomment-1982803504. 

I'd be happy to switch to a simpler pair of words, but there doesn't seem to be any official or popular terminology for this. I might add a glossary section to explain some of the opaque terminology I use within the article.

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+- Explain concepts that can help resolve recurring issues that aren't preventable with compliance alone.
+- Serve as a reference during code review to encourage preferred approaches, minimize redundant discussion, and educate contributors about established practices.
+
+This article does not aim to:
+
+- Be a comprehensive resource covering all major aspects of syntax and style.
+- Redundantly list or explain conventions that are enforceable through linters or static rulesets.
+- Endorse its contents as being appropriate or useful for any TypeScript codebase outside of our own.
+
+## Types
```
Gudahtt, line 15: Great idea to summarize this at the beginning, I found this helpful

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+
+- In general, using `any` in this context is not harmful in the same way that it is in other contexts, as the `any` types only are not directly assigned to any specific variable, and only function as constraints.
+- That said, more specific constraints provide better type safety and intellisense, and should be preferred wherever possible.
+
+##### `any` may be acceptable to use to type the error property in a catch block
+
+- `catch` only accepts `any` and `unknown` as the error type.
+- Recommended: Use `unknown` with type guards like `isJsonRpcError`.
+- Avoid typing an error object with `any` if it is passed on to be used elsewhere instead of just being thrown, as the `any` type will infect the downstream code.
+
```
MajorLift, line 581:
```suggestion
```

docs/typescript.md
MajorLift, line 584:
```suggestion
```

docs/typescript.md
Gudahtt: Ah sorry, to clarify, I do understand what the terms mean. Just not how it's used in this context. Does "When typing the assigned" mean "When writing a type annotation"?

When this says "`unknown` is unable to replace `any`", does this mean that changing a type annotation from `any` to `unknown` can introduce errors?

I get that this situation is about explaining when `never` is an appropriate replacement for `any`, but I'm not grasping what the specific examples/scenarios are here.

docs/typescript.md
MajorLift: Applied here: https://github.com/MetaMask/contributor-docs/pull/74/commits/5d81b66240fc5f299e943da9afb8bea27a4a51f1

docs/typescript.md
MajorLift: Applied here https://github.com/MetaMask/contributor-docs/pull/74/commits/5d81b66240fc5f299e943da9afb8bea27a4a51f1

docs/typescript.md
MajorLift: Applied here https://github.com/MetaMask/contributor-docs/pull/74/commits/5d81b66240fc5f299e943da9afb8bea27a4a51f1

docs/typescript.md
MajorLift: Applied here https://github.com/MetaMask/contributor-docs/pull/74/commits/5d81b66240fc5f299e943da9afb8bea27a4a51f1

docs/typescript.md
```diff
@@ -1,14 +1,590 @@
...
+// @ts-expect-error Intentionally testing invalid state
+
+// @ts-expect-error We are intentionally passing bad input.
+```
+
+##### If accompanied by a TODO comment, `@ts-expect-error` is acceptable to use for marking errors that have clear plans of being resolved
+
+<!-- TODO: Add example -->
+
+#### Avoid `any`
```
Gudahtt, line 407: It looks like this is currently in the "Compiler Directives" section. Was this meant to be its own section?

docs/typescript.md
MajorLift: This is related to a specific error I encountered on the `any` ticket for rate-limit-controller, and the example is [here](https://github.com/MetaMask/contributor-docs/pull/74/files/e6114410c2115c6783e0ccece94cbbaa336df6db#diff-ae03a9f2eb638a25c1d8d7ba853c78220f1ed44a19ad97a8d780cbc0c00a4260R528-R541). I'm realizing the order of presentation might be a little confusing on these examples hmm...

Gudahtt: Oh I see, sorry no I did see the example but for some reason I assumed it was an "artificial" situation written to demonstrate this point. But I see that it is a real example.

I thought this guideline was about adding or deleting properties in a way that is disallowed by the types. But I guess it's meant for situations where the type allows it, but TypeScript doesn't properly "collapse" the generic to the specific example the code applies to.

Gudahtt: I'm curious about what the limitation is here, do you know if it's the same as what this bug describes? https://github.com/microsoft/TypeScript/issues/36631
Is there another workaround that we could consider here?

But more generally, do you think we should encourage `@ts-expect-error` here or a type assertion, rather than `any`? At least it would be less "infectious".

docs/typescript.md
MajorLift: Designating `any` as a compiler directive was intentional, just to highlight that it's not an actual type. 

The hierarchy of headings is a little iffy. I currently have "Avoid something" and "Acceptable usages of something" consistently at the same level (h4), because pushing down the actual entries (h6) an additional level would make them too small to read.

docs/typescript.md
MajorLift: Yes `unknown` and `any` are only interchangeable if they're typing the assignee. 

If they're typing the assigned, it's likely that the `any` was used in the first place because the assigned type is too wide to fit into the assignee type "slot", resulting in an error. Using `unknown`, the widest type, as the substitute probably won't help. there. 

`never` will at least fit into the assignee, since it's assignable to every type.

Gudahtt: Hmm, I'm still not following how these points help with understanding when `never` can be used to replace `any`. If there is a broader point here, I am missing it.

docs/typescript.md
Gudahtt: Hmm, I had never heard of `any` being referred to as a "compiler directive". I took "compiler directive" to mean the triple-slash directives. Even if this is technically true on some level, it doesn't seem like the most intuitive way to organize the document.

MajorLift: Hmm I can see that "compiler directives" could be confusing. What do you think about naming the last section "Escape Hatches" instead? That could cover `@ts-expect-error`, `@ts-ignore`, `any`, in order of preference. 

The `@ts-ignore` section could only include a brief paragraph comparing it to `any`: The two are functionally identical, except `any` is much worse, because it's like applying `@ts-ignore` to every occurrence of the assignee variable throughout the codebase and even in downstream modules. That and the fact that `@ts-ignore` applies to entire lines that could contain multiple unrelated errors are the only differences.

I think this would provide better context for the `any` section that follows so it doesn't feel out of place.

Gudahtt: "Escape hatches" sounds like a great description of what these are

docs/typescript.md
MajorLift: Well our `ActionConstraint['handler']` going from `(...args: any[]) => unknown` to `(...args: never[]) => unknown` is a good example, although function param contravariance is a bit of a special case.

Other than that, if the any-typed variable is going to be assigned to something else e.g. passed into another function as an argument, fixing it to `unknown` will not work (unless the assignee is also `unknown`), while `never` will always work. Here's an example:

```ts
// No error when `arg1` is `any` or `never`. 
// If `arg1` is `unknown`:
// Error: Argument of type 'unknown' is not assignable to parameter of type 'string'.(2345)
function f1(arg1: ???) {
  f2(arg1)
}
function f2(arg2: string) { ... }
```
https://tsplay.dev/mpodxN

docs/typescript.md
```diff
@@ -1,14 +1,597 @@
...
+  'NetworkController:getProviderConfig',
+).chainId; // Type '`0x${string}`'
+
+const BUILT_IN_NETWORKS = {
+  mainnet: '0x1',
+  sepolia: '0xaa36a7',
+} as const; // Type { readonly mainnet: '0x1'; readonly sepolia: '0xaa36a7'; }
+```
+
+```typescript
```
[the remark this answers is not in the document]
MajorLift: Thanks for pointing this out! I added headings on the examples which should resolve this: https://github.com/MetaMask/contributor-docs/pull/74/commits/c7be62df8a8509c35ae46fd5c3e5c508555f9c89

docs/typescript.md
MajorLift: Fixed: https://github.com/MetaMask/contributor-docs/pull/74/commits/c7be62df8a8509c35ae46fd5c3e5c508555f9c89

docs/typescript.md
MajorLift: I think this is a different issue. In that bug, the index signature is not being recognized as valid, which is why the workaround works by constraining the type of `K`. 

In this example, the index signature is valid, but the issue is it being used in a write operation on a generic type.

On the subject of what our recommendation should be...

`as any` doesn't seem to be "infectious" like `any`. It only applies to the target variable and not to other instances of the same variable. I should add a sentence about this, but without necessarily encouraging its use.

`@ts-expect-error` is probably the best option here. There might be an argument for conceptually distinguishing `@ts-expect-error` and `any` use cases. But from a pragmatic viewpoint, I think preferring `@ts-expect-error` in all cases makes sense because of how it alerts us when the error it's suppressing is no longer there.

I'll move this over to the `@ts-expect-error` section if that sounds ok.

Gudahtt: Sounds good! That makes sense to me

docs/typescript.md
mcmire: I have similar thoughts. I understand that people may have questions about when it is okay to use `as`, and because of this, perhaps this section is intended to be informational rather than advisory. But, I worry that people may cite this section as evidence that it is okay to do this, especially since we say that this is "acceptable", and to avoid this I feel like we need to be more guarded in the language.

I think it would be more helpful to make this a proper guideline and then to use specific language which clarifies _when_ it is okay. Perhaps instead of "`as` is acceptable to use for preventing or fixing `any` usage" we say "Prefer type assertions over disabling type-checking" or something like that?

Besides this, I'm curious if there is a better example, since we don't want people testing private methods to begin with.

docs/typescript.md
```diff
@@ -1,14 +1,644 @@
...
+```
+
+✅ Compiler specifies that the target type should be `keyof NftController`
+
+```typescript
+sinon.stub(nftController, 'getNftInformation' as keyof typeof nftController);
+```
+
+- Even an assertion to a wrong type still allows the compiler to show us warnings and errors as the code changes, and is therefore preferrable to the dangerous radio silence enforced by `any`.
+- For type assertions to an incompatible shape, use `as unknown as` as a last resort rather than `any` or `as any`.
```
mcmire, line 352: Should we provide an example for this? Or — maybe _this_ should be the example? I feel like there are legitimate cases for this where there are no other workarounds.

docs/typescript.md
Gudahtt: Hmm. That is a good example of why using `never` as a type annotation is problematic, but is there ever a reason to consider `never` here?

I've only seen `never` used personally in conditional types, or in generics. In conditional types its use is rather obvious, and could never be confused with `any`. So are generics the last case where we might expect to replace `any` with `never`? Or is there another case that I'm missing

docs/typescript.md
```diff
@@ -1,14 +1,644 @@
...
+sinon.stub(nftController, 'getNftInformation' as any);
+```
+
+✅ Compiler specifies that the target type should be `keyof NftController`
+
+```typescript
+sinon.stub(nftController, 'getNftInformation' as keyof typeof nftController);
+```
+
+- Even an assertion to a wrong type still allows the compiler to show us warnings and errors as the code changes, and is therefore preferrable to the dangerous radio silence enforced by `any`.
```
mcmire, line 351: Should we move to the list of points above the example? It seems more like to be noticed.

docs/typescript.md
mcmire: Perhaps we want to use `as const` as an example?

docs/typescript.md
mcmire: I know that this guideline came out of a discussion that you and I had but I'm not sure I agree with it. If we are creating an object in a test and we need to satisfy a type, I feel that we should prefer a full representation of that type rather than a partial representation. Going further, though, if this results in tests that are difficult to maintain or hard to read, then this is a smell, and we should see if we can have the thing we are testing take a smaller portion of that object rather than the whole thing. I think that is a better route than "lying" in tests. Thoughts?

docs/typescript.md
```diff
@@ -1,14 +1,644 @@
...
+
+<!-- TODO: Add examples -->
+
+### Compiler Directives
+
+TypeScript provides several directive comments that can be used to suppress TypeScript compiler errors. Using these to ignore typing issues is dangerous and reduces the overall effectiveness of TypeScript.
+
+Of these, we will discuss the use cases and pitfalls of `@ts-expect-error` and `any`.
+
+`@ts-ignore`, `@ts-nocheck`, `@ts-check` are disabled by ESLint rules.
```
mcmire, line 399: Do we want to link to our `@metamask/eslint-config-typescript` package where this is disabled? Perhaps not the exact line in case it ever changes, but just the package itself: https://github.com/MetaMask/eslint-config/tree/main/packages/typescript

docs/typescript.md
```diff
@@ -1,14 +1,644 @@
...
+  | ((payload_0: any, payload_1: Patch[]) => void);
+
+function returnsAny(): any {
+  return { a: 1, b: true, c: 'c' };
+}
+// Types of a, b, c are all `any`
+const { a, b, c } = returnsAny();
+```
+
+- `any` infects all surrounding and downstream code with its directive to suppress errors. This is the most dangerous characteristic of `any`, as it causes the encroachment of unsafe code with no guarantees about type safety or runtime behavior.
```
mcmire, line 476: Nit: Should this be a bullet point, if the next paragraph is not? Or should the next paragraph also be a bullet point?

docs/typescript.md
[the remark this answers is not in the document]
mcmire: I agree that this section is a bit technical. It seems that we could make this a bit more concise. I'll see if I can think of some suggestions, but if not then we can always improve this later.

mcmire (commented):
Sorry for the very late review on this. I took a pass and had some suggestions. I think most are minor. If there are any that require discussion, I tried to make some suggestions but we can always come back later and add polish. My review was kind of low-level so I'll try to do a high-level review later today.

docs/typescript.md
```diff
@@ -2,13 +2,473 @@
...
+- `any` usage is often motivated by a need to find a placeholder type that could be anything. `unknown` is the most likely type-safe substitute for `any` in these cases.
+
+###### `never`
+
+- `never` is the universal subtype i.e. the narrowest possible type, equivalent to the null set(∅).
+- `never` is assignable to every type, but the only type that is assignable to `never` is `never`.
+- When typing the _assigned_:
+  - `unknown` is unable to replace `any`, as `unknown` is only assignable to `unknown`.
+  - The type of the _assigned_ must be a subtype of the _assignee_.
+  - `never` is worth trying, as it is the universal subtype and assignable to all types.
```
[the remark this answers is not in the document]
mcmire: This suggestion was from a previous review. It is now outdated and is somewhat complete but is hopefully somewhat useful. I will leave a more useful suggestion on this later.

docs/typescript.md
```diff
@@ -1,14 +1,644 @@
...
+### Type Annotations
+
+An explicit type annotation is acceptable for overriding an inferred type if...
+
+1. It can further narrow an inferred type, thus supplying type information that the compiler cannot infer or access.
+2. It is being used to enforce a type constraint rather than assign a type definition.
+
+##### Use the `satisfies` operator to enforce a type constraint or to validate the assigned type
+
+TypeScript provides the `satisfies` operator for constraining or validating a type. It is able to both enforce a type constraint and further narrow the assigned type through inference.
```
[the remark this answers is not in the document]
MajorLift: This seems helpful and unobtrusive. Applied!

docs/typescript.md
```diff
@@ -1,14 +1,644 @@
...
+
+##### Use type annotations if it will prevent or remove type assertions
+
+Type annotations are more responsive to code drift than assertions. If the assignee's type becomes incompatible with the assigned type annotation, the compiler will raise a type error, whereas in most cases a type assertion will still suppress the error.
+
+When the compiler is in doubt, an annotation will nudge it towards relying on type inference, while an assertion will force it to accept the user-supplied type.
+
+<!-- TODO: Add example -->
+
+##### When instantiating an empty container type, provide a type annotation
```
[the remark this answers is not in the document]
MajorLift: Hmm my intention was to talk about instantiating a value. Maybe this is more clear?

```suggestion
##### When instantiating an empty composite data-type value, provide a type annotation 
```

docs/typescript.md
MajorLift: Yes, it shouldn't be too hard to find examples for this. Adding a TODO for now.

```suggestion
- For type assertions to an incompatible shape, use `as unknown as` as a last resort rather than `any` or `as any`.
<!-- TODO: Add example -->
```

docs/typescript.md
MajorLift: The example here was intended to only apply to the specific bullet point it's under. I think moving that bullet point into its own entry as per https://github.com/MetaMask/contributor-docs/pull/74/files#r1523858021 will resolve this as well.

docs/typescript.md
MajorLift: This line is part of the list in lines 457-459, with the example above only applying to line 459. Maybe a better example format with inline-html might make lists that include examples easier to parse? Otherwise, I'm not sure how to handle examples that are inserted into lists or otherwise don't quite fall under a dedicated entry.

docs/typescript.md
```diff
@@ -2,13 +2,473 @@
...
+
+Some generic types use `any` as a default generic argument. This can silently introduce an `any` type into the code, causing unexpected behavior and suppressing useful errors.
+
+🚫
+
+```typescript
+const NETWORKS = new Map({
+  mainnet: '0x1',
+  goerli: '0x5',
+}); // Type 'Map<any, any>'
```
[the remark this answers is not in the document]
MajorLift: Ah looks like `Map<any, any>` only happens if it's instantiated as an empty collection. Since this case is already covered in the "When instantiating an empty container type, provide a type annotation" entry, I'm good with only leaving the `jest.fn` example here. If we find any other good examples we can add them later.

docs/typescript.md
```diff
@@ -1,14 +1,644 @@
...
+```
+
+✅
+
+```typescript
+const NETWORKS = new Map<string, `0x${string}`>({
+  mainnet: '0x1',
+  goerli: '0x5',
+}); // Type 'Map<string, `0x${string}`>'
+
```
MajorLift, line 531:
```suggestion
```

docs/typescript.md
MajorLift: I think it would be clearer to rewrite this so that the distinction is between cases where `unknown` works as a substitute for `any` vs. cases where it doesn't. 

I agree that in the second case, it's not especially helpful to single out `never` as the first option to try just because it will almost always work.

In terms of `never` use cases, I have a draft for another entry: ["Prefer `as never` for empty types"](https://github.com/MetaMask/contributor-docs/issues/69#issuecomment-1910430960). But that isn't about replacing `any` so it doesn't apply here. 

It seems like the only `never` use case left to cover here is as a possible replacement for `any` in generic arguments.

docs/typescript.md
```diff
@@ -1,14 +1,637 @@
...
+```typescript
+function f(x: SomeInterface | SomeOtherInterface) {
+  if (isSomeInterface(x)) {
+    console.log(x.name); // Type of x: 'SomeInterface'. Type of x.name: 'string'.
+  }
+}
+```
+
+###### Example (f7ff4b0d-e5e9-4568-b916-5153ddd2095b)
```
MajorLift, line 277: Add note that this example will become obsolete in TypeScript v5.5: https://github.com/microsoft/TypeScript/pull/57465

mcmire: Yeah I saw that! So cool.

docs/typescript.md
```diff
@@ -1,14 +1,637 @@
 # TypeScript Guidelines
 
-These guidelines specifically apply to TypeScript.
+## Introduction
 
-## Provide explicit return types for functions/methods
+These guidelines apply to TypeScript code in MetaMask repos.
+
```
mcmire, line 6: This sentence was originally an analogue to the sentence at the beginning of the JavaScript guidelines, but I had been meaning to remove it as it's kind of redundant. Thoughts on removing?

``` suggestion
```

docs/typescript.md
```diff
@@ -1,14 +1,590 @@
...
+### Type Inference
+
+TypeScript is very good at inferring types. Explicit type annotations and assertions are the exception rather than the rule in a well-managed TypeScript codebase.
+
+Some fundamental type information must always be supplied by the user, such as function and class signatures, interfaces for interacting with external entities or data types, and types that express the domain model of the codebase.
+
+However, for most types, inference should be preferred over annotations and assertions.
+
+##### Prefer type inference over annotations and assertions
```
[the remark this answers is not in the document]
mcmire: Hmm. I'm not sure I see it this way. I feel like the guidelines are easy to find in the navigation even if not every one is intended at the same level. I feel like the jump in header levels is more likely to be seen as an inconsistency.

docs/typescript.md
MajorLift: Applied here: https://github.com/MetaMask/contributor-docs/pull/74/commits/03bdb01d4210fa5fa0e24e92c672f3e2bdf6f8c0

docs/typescript.md
MajorLift: I ended up removing the "assignment to generics" example and creating a new "Prefer `as unknown as` over `as any`" entry: https://github.com/MetaMask/contributor-docs/pull/74/commits/aadb1051362540afe53be54acb6d45d718dc2470.

I'm not sure if the generics assignment is the best example to present as a `@ts-expect-error` use case. I'll bring it back if I can find a good slot for it.

docs/typescript.md
MajorLift: Applied here: https://github.com/MetaMask/contributor-docs/pull/74/commits/6f229675d64adb4fcc2452c89c5b86c7f4a1f604

docs/typescript.md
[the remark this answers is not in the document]
MajorLift: Might be resolved by this: https://github.com/MetaMask/contributor-docs/pull/74/commits/03bdb01d4210fa5fa0e24e92c672f3e2bdf6f8c0

docs/typescript.md
MajorLift: Removed this section: https://github.com/MetaMask/contributor-docs/pull/74/commits/f627cbcd3297d22d46a838cc2e4f2defdcc2f174, and moved content directly under "Type Annotation" heading: https://github.com/MetaMask/contributor-docs/pull/74/commits/37773897aad289d77c9f9a9f353a157525dcbc7a

docs/typescript.md
MajorLift: Removed this section: https://github.com/MetaMask/contributor-docs/pull/74/commits/f627cbcd3297d22d46a838cc2e4f2defdcc2f174, and moved content directly under "Type Annotation" heading: https://github.com/MetaMask/contributor-docs/pull/74/commits/37773897aad289d77c9f9a9f353a157525dcbc7a

docs/typescript.md
MajorLift: Applied: https://github.com/MetaMask/contributor-docs/pull/74/commits/58ed3787ace1c5ed2e41ecbb0c900f94eb3070b1

mcmire (commented):
Made a full pass and had some more comments on ways we could tighten this doc. I feel like we're getting closer.

docs/typescript.md
mcmire: The new changes go into more detail, so I'm good on this.

docs/typescript.md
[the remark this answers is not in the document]
mcmire: Yes, the new changes are better!

docs/typescript.md
mcmire: I am okay with leaving the example for now and improving it later (given that it _is_ based on a real-world example), but I'm still curious whether it makes sense to rename the header.

docs/typescript.md
mcmire: For what it's worth, the new changes are more approachable to me.

docs/typescript.md
mcmire: I like this as well!

docs/typescript.md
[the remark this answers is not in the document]
mcmire: (Brought out into its own comment for visibility) This is minor, as it doesn't make it more difficult to read, so feel free to ignore for now.

mcmire (approved):
Lots of good stuff to chew on here! I skimmed through it one more time. This gets me thinking that it may be valuable to have two different kinds of documents in the future, one that's purely "do"s and "don't"s, and another that is a bit more expositional. We can decide on that in the future though. Nice work!

bdstar81 (approved):

PR #80, opened 2024-02-28
add document for typescript guidelines

DDDDDanica:
Provide [basic guideline for typescript](https://github.com/MetaMask/contributor-docs/blob/0a2b7c426d238adefadac7cc51eee20875d25785/docs/typescript.md)

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -1,14 +1,14 @@
-# **Typescript guidelines**
+# Typescript Guidelines
 
 **_Purpose_**: Establish best practices for writing robust, maintainable, and efficient TypeScript code.
 
 **_Audience_**: Internal and external developers using TypeScript in Metamask.
 
-## **Interfaces**
+## **Type**
 
 By defining expected object structures with interfaces, TypeScript promotes developer intent and unlocks powerful IDE features like accurate completion and context-aware navigation.
 
-### **Enumerations**
+### **Enums**
 
 TypeScript offers several tools for crafting clear data definitions, with enumerations and unions standing as popular choices.
 
@@ -16,9 +16,50 @@ TypeScript offers several tools for crafting clear data definitions, with enumer
 
 Inevitably you will want to refer to the values of a union type somewhere (perhaps as the argument to a function). You can of course just use a literal which represents a member of that union — but if you have an enum, then all of the values are special, and any time you use a value then anyone can see where that value comes from.
 
+🚫
+
+```typescript
+type UserRole = 'admin' | 'editor' | 'subscriber';
+```
+
+✅
+
+```typescript
+enum AccountType {
+  Admin = 'admin',
+  User = 'user',
+  Guest = 'guest',
+}
+```
+
 **Don't use numeric enums**
 
 Numeric enums are misleading because it creates a reverse mapping from value to property name, and when using `Object.values` to access member names, it will return the numerical values instead of the member names, potentially causing unexpected behavior.
+🚫
+
+```typescript
+enum Direction {
+  Up = 0,
+  Down = 1,
+  Left = 2,
+  Right = 3,
+}
+
+const directions = Object.values(Direction); // [0, 1, 2, 3]
+```
+
+✅
+
+```typescript
+enum Direction {
+  Up = 'Up',
+  Down = 'Down',
+  Left = 'Left',
+  Right = 'Right',
+}
+
+const directions = Object.values(Direction); // ["Up", "Down", "Left", "Right"]
+```
 
 ## **Types**
 
@@ -28,14 +69,12 @@ It may not be enough just to have a type or a function take another type — you
 
 ```typescript
 // before
-function createExampleMiddleware(
-  exampleParam,
-): JsonRpcMiddleware<JsonRpcParams, Json>;
+function createExampleMiddleware<Params, Result>(exampleParam);
 // after
 function createExampleMiddleware<
-  Params extends JsonRpcParams = JsonRpcParmas,
-  Result extends Json = Json,
->(exampleParam): JsonRpcMiddleware<Params, Result>;
+  Params extends JsonRpcParams,
+  Result extends Json,
+>(exampleParam);
 ```
 
 #### Use `Omit` to reduce requirements
@@ -77,7 +116,7 @@ const cartPayload: singleItemPayload[] = [];
 
 #### Avoid type assertions, or explain their purpose whenever necessary.
 
-Type assertions inform TypeScript about a variable's actual type, using the `as` syntax or the angle bracket `<Type>` syntax.
+Type assertions inform TypeScript about a variable's actual type, using the `as` syntax.
 While type assertions offer flexibility, they can bypass TypeScript's safety net. This means the compiler won't catch potential type mismatches, leaving you responsible for ensuring correctness.
 
 While it's generally recommended to avoid type assertions in TypeScript, here are two common scenarios where they might be necessary:
@@ -109,7 +148,7 @@ console.log(youngestUser); // { name: "...", age: maxAge }
 2. Iterating over Enum Keys:
    When iterating over enum keys using `Object.keys`, the resulting array is of type `string[]`. However, if you want to access enum member values directly, you might need a type assertion.
 
-```typescript=
+```typescript
 enum Direction {
   Up,
   Down,
@@ -119,8 +158,8 @@ enum Direction {
 
 const directions: Direction[] = [Direction.Up, Direction.Down];
 
-for (const key in directions) {
-  const direction = directions[key] as Direction; // Assert to access enum values.
+for (const key of directions) {
+  const direction = directions[key];
   console.log(direction); // Output: Direction.Up, Direction.Down
 }
 ```
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -1,14 +1,225 @@
-# TypeScript Guidelines
+# **Typescript guidelines**
 
-These guidelines specifically apply to TypeScript.
+***Purpose***: Establish best practices for writing robust, maintainable, and efficient TypeScript code.
 
-## Provide explicit return types for functions/methods
+***Audience***: Internal and external developers using TypeScript in Metamask.
 
+
+## **Interfaces**
+By defining expected object structures with interfaces, TypeScript promotes developer intent and unlocks powerful IDE features like accurate completion and context-aware navigation.
+
+### **Enumerations**
+TypeScript offers several tools for crafting clear data definitions, with enumerations and unions standing as popular choices.
+
+**Consider using enums over union types for situations with a fixed set of known values.**
+
+Inevitably you will want to refer to the values of a union type somewhere (perhaps as the argument to a function). You can of course just use a literal which represents a member of that union — but if you have an enum, then all of the values are special, and any time you use a value then anyone can see where that value comes from.
+
+**Don't use numeric enums**
+
+Numeric enums are misleading because it creates a reverse mapping from value to property name, and when using `Object.values` to access member names, it will return the numerical values instead of the member names, potentially causing unexpected behavior.
+
+
+
+## **Types**
+#### Constrain generic types if necessary
+It may not be enough just to have a type or a function take another type — you might have to constrain it if it's not allowed to be anything (e.g. extends Json)
+
+```typescript
+// before
+function createExampleMiddleware(exampleParam): JsonRpcMiddleware<JsonRpcParams, Json>
+// after
+function createExampleMiddleware<
+  Params extends JsonRpcParams = JsonRpcParmas, 
+  Result extends Json = Json
+>(exampleParam): JsonRpcMiddleware<Params, Result>
+```
+
+#### Use `Partial` to type the accumulating variable in `reduce`
+When using reduce in TypeScript, the accumulating variable's type can be difficult to define, especially when dealing with optional properties or incomplete objects. In such cases, using Partial allows you to start with an empty object and gradually add optional properties as needed. Here's an example:
+
+``` typescript
+interface User {
+  name: string;
+  age?: number; // optional property
+  email: string;
+}
+
+// Partial<User> would be:
+type PartialUser = {
+  name?: string;
+  age?: number;
+  email?: string;
+}
+
+const users: User[] = [
+  { name: "Alice", age: 25, email: "alice@example.com" },
+  { name: "Bob", age: 30, email: "bob@example.com" },
+];
+
+const oldestUser = users.reduce((acc, user) => {
+  // acc starts as Partial<User> (empty object)
+  if (user.age && (acc.age === undefined || user.age > acc.age)) {
+    acc.age = user.age;
+    acc.name = user.name; // add additional properties as needed
+  }
+  return acc;
+}, {} as Partial<User>); // initialize with empty object
+
+console.log(oldestUser); // { name: "Bob", age: 30 }
+```
+
+
+#### Use `Omit` to reduce requirements
+`Omit<T, K>` takes two generic types: `T` representing the original object type and `K` representing the property keys you want to remove. It returns a new type that has all the properties of T except for the ones specified in K. Here are some cases to use omit:
+- Removing Unnecessary Properties:
+  Imagine you have a user interface with optional email and phone number fields. However, your API call only cares about the `username`. You can use Omit to reduce the required properties:
+``` typescript
+interface User {
+  username: string;
+  email?: string;
+  phoneNumber?: string;
+}
+
+// Type for API call payload
+type ApiPayload = Omit<User, "email" | "phoneNumber">;
+
+const payload: ApiPayload = { username: "johndoe" };
+// Now `payload` only has the `username` property, satisfying the API requirements.
+```
+
+- Conditional Omission:
+  Sometimes, you might want to remove properties based on a condition. `Omit` can still be helpful:
+```typescript
+interface CartItem {
+  productId: number;
+  quantity: number;
+  color?: string; // Optional color
+
+// Omit color if quantity is 1
+const singleItemPayload = Omit<CartItem, "color" extends string ? "color" : never>;
+
+// Omit color for all items if quantity is always 1
+const cartPayload: singleItemPayload[] = [];
+```
+
+#### When augmenting an object type with `&`, remember to apply `Omit` to any overlapping property keys.
+
+* `&` does not automatically overwrite overlapping entries like the spread (...) operator does: https://tsplay.dev/NDJ48W
+```typescript
+type Original = {
+  a: string;
+  b: number;
+  c: boolean;
+}
+type Augment = {
+  c: 'hello'
+}
+type NoOmit = Original & Augment // never
+type UseOmit = Omit<Original, 'c'> & Augment // { a: string; b: number; c: 'hello'; }
+```
+
+* One edge case to be especially careful of is if an optional property is being overwritten. Unlike the above case, this will fail silently: https://tsplay.dev/m0OBxN
+```typescript
+type Original = {
+  a: string;
+  b: number;
+  c?: boolean;
+}
+type Augment = {
+  c?: 'hello'
+}
+type NoOmit = Original & Augment // { a: string; b: number; c?: undefined; }
+```
+
+
+#### Don't use `any`
+
+* `any` doesn't actually represent "any type"; if you declare a value as `any`, it tells TypeScript to disable typechecking for that value anywhere it appears. This means the compiler won't warn you about potential type mismatches or errors, leading to hidden issues that might surface later in development or production.
+* `any` is dangerous because it infects all surrounding and downstream code. For instance, if you have a function that is declared to return `any` which actually returns an object, all properties of that object will be `any`.
+* Need to truly represent something as "any value"? Use `unknown` instead.
+
+#### Avoid type assertions, or explain their purpose whenever necessary.
+Type assertions inform TypeScript about a variable's actual type, using the `as` syntax or the angle bracket ``<Type>`` syntax.
+While type assertions offer flexibility, they can bypass TypeScript's safety net. This means the compiler won't catch potential type mismatches, leaving you responsible for ensuring correctness.
+
+
+While it's generally recommended to avoid type assertions in TypeScript, here are two common scenarios where they might be necessary:
+
+1. Using reduce:
+
+Imagine you have an array of User objects and want to find the youngest user using reduce. You start with an empty accumulator `({})`, but the final result should be a User object.
+```typescript
+interface User {
+  name: string;
+  age: number;
+}
+
+const users: User[] = [
+  // ... user data
+];
+
+const youngestUser = users.reduce((acc, user) => {
+  if (!acc.age || user.age < acc.age) {
+    acc = user; // Here, `acc` might be undefined initially.
+  }
+  return acc;
+}, {} as User); // Assert `{}` as `User` to satisfy the return type.
+
+console.log(youngestUser); // { name: "...", age: maxAge }
+```
+
+2. Iterating over Enum Keys:
+   When iterating over enum keys using `Object.keys`, the resulting array is of type `string[]`. However, if you want to access enum member values directly, you might need a type assertion.
+```typescript=
+enum Direction {
+  Up,
+  Down,
+  Left,
+  Right,
+}
+
+const directions: Direction[] = [Direction.Up, Direction.Down];
+
+for (const key in directions) {
+  const direction = directions[key] as Direction; // Assert to access enum values.
+  console.log(direction); // Output: Direction.Up, Direction.Down
+}
+```
+
+#### Avoid type annotations
+Type annotations act like contracts with the compiler, ensuring a variable sticks to a specific data type its entire life.
+
+In rare cases, using type annotations with `reduce` can be beneficial. It helps tp clarify complex accumulator types, enhance readability and ensure the compiler understands its intended use.
+
+If you need to create a data structure with specific values, but ensure that it's declared with a more abstract type. Create a function that takes an argument of the type you want, then use it to build the object.
+```typescript
+interface Person {
+  name: string;
+  age: number;
+  occupation?: string; // Optional property
+}
+
+// Function with type annotations:
+function createPerson<T extends { name: string; age: number }>(data: T): Person {
+  // Infer types from the data argument
+  const { name, age, occupation = "Unknown" } = data;
+
+  // Ensure properties adhere to the Person interface
+  return { name, age, occupation };
+}
+
+// Usage:
+const person1 = createPerson({ name: "Alice", age: 30 }); // { name: "Alice", age: 30, occupation: "Unknown" }
+const person2 = createPerson({ name: "Bob", age: 25, occupation: "Developer" }); // { name: "Bob", age: 25, occupation: "Developer" }
+```
+TypeScript is planning to introduce the `satisfies` keyword, which will provide a more flexible way to declare that a data structure conforms to a specific type without sacrificing readability or maintainability. However, this feature is not yet available.
+
+#### Provide explicit return types for functions/methods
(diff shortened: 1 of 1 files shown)


docs/typescript.md
```diff
@@ -1,14 +1,225 @@
...
+
+## **Interfaces**
+By defining expected object structures with interfaces, TypeScript promotes developer intent and unlocks powerful IDE features like accurate completion and context-aware navigation.
+
+### **Enumerations**
+TypeScript offers several tools for crafting clear data definitions, with enumerations and unions standing as popular choices.
+
+**Consider using enums over union types for situations with a fixed set of known values.**
+
+Inevitably you will want to refer to the values of a union type somewhere (perhaps as the argument to a function). You can of course just use a literal which represents a member of that union — but if you have an enum, then all of the values are special, and any time you use a value then anyone can see where that value comes from.
```
NicolasMassart, line 16: Could you provide examples for this?

docs/typescript.md
```diff
@@ -1,14 +1,225 @@
...
+
+**Don't use numeric enums**
+
+Numeric enums are misleading because it creates a reverse mapping from value to property name, and when using `Object.values` to access member names, it will return the numerical values instead of the member names, potentially causing unexpected behavior.
```
NicolasMassart, line 20: Could you provide examples for this?

docs/typescript.md
```diff
@@ -39,3 +249,230 @@ async function removeAccount(address: Hex): Promise<KeyringControllerState> {
...
+const walletName = 'metamask';
+this.messagingSystem.registerActionHandler(
+      `${name}:call`,
+      ...
+);
+```
+
+
+## **Language features**
+#### Prefer the nullish coalescing operator `??`
```
NicolasMassart, line 383: Could you provide examples for this?

docs/typescript.md
```diff
@@ -39,3 +249,230 @@ async function removeAccount(address: Hex): Promise<KeyringControllerState> {
...
+* `??` guarantees that the left-hand operand is nullish but not falsy, which is both safer and reduces cognitive overhead for reviewers.
+* Using `||` leads reviewers to assume that ?? was intentionally avoided, forcing them to consider cases where the left-hand operand is falsy. This should never happen if the left-hand operand is guaranteed to be nullish and not falsy.
+
+#### Prefer negation (`!`) over strict equality checks for `undefined`, but not for `null`.
```
NicolasMassart, line 387: Could you provide examples for this?

docs/typescript.md
```diff
@@ -39,3 +249,230 @@ async function removeAccount(address: Hex): Promise<KeyringControllerState> {
...
+
+In general, array types of arbitrary length are not assignable to tuple types that have fixed types designated to some or all of its index positions (the array is wider than the tuple since it has less constraints).
+
+This means that there's a whole subcategory of functions that `(...args: never[]) => unknown` doesn't cover: functions that have a finite number of parameters that are strictly ordered and typed, while also accepting an arbitrary number of "rest" parameters. The spread argument list of such a function evaluates as a variadic tuple type with a minimum fixed length, instead of a freely expandable (or contractable) array type.
+
+`(...args: never[]) => unknown` covers functions for which all of the params can be typed together as a group e.g. `(...args: (string | number)[]) => T`, or it has a fixed number of params e.g. `(a: string, b: number) => T`. But to cover cases like `(a: string, b: number, ...rest: unknown[]) => T`, we need another top type with an even narrower type for the params: `(...args: never) => unknown`.
+
+
+#### Avoid implicit, hard-coded generic parameters that are inaccessible from the outermost user-facing type.
```
NicolasMassart, line 471: Could you provide examples for this?

NicolasMassart (commented):
Nice work!
Will definitely help us.
I just pointed places where, as a new TS dev, I could be helped by some examples.

MajorLift (changes requested):
Hi, great work putting this together! 

I noticed that you included my work in this (https://github.com/MetaMask/contributor-docs/issues/69#issuecomment-1910430960). These entries are still WIP, and not intended to be published yet. I'd appreciate if you could remove them. I'll submit them when they're ready.

Also, a heads-up that my PR for https://github.com/MetaMask/contributor-docs/issues/47 is almost ready for review, so any content here relating to `any`, type assertions, type annotations may be edited to remove duplication and make the doc read coherently.

docs/typescript.md
```diff
@@ -1,14 +1,225 @@
...
+  // acc starts as Partial<User> (empty object)
+  if (user.age && (acc.age === undefined || user.age > acc.age)) {
+    acc.age = user.age;
+    acc.name = user.name; // add additional properties as needed
+  }
+  return acc;
+}, {} as Partial<User>); // initialize with empty object
+
+console.log(oldestUser); // { name: "Bob", age: 30 }
+```
```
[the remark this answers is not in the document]
MajorLift: This also conflicts directly with [lines 412-439](https://github.com/MetaMask/contributor-docs/pull/80#discussion_r1506095370).

docs/typescript.md
```diff
@@ -1,14 +1,225 @@
...
+
+const youngestUser = users.reduce((acc, user) => {
+  if (!acc.age || user.age < acc.age) {
+    acc = user; // Here, `acc` might be undefined initially.
+  }
+  return acc;
+}, {} as User); // Assert `{}` as `User` to satisfy the return type.
+
+console.log(youngestUser); // { name: "...", age: maxAge }
+```
```
MajorLift, line 170: There are four different entries providing conflicting guidelines for `reduce` usage. 

This highlighted portion aligns with this entry:
- https://github.com/MetaMask/contributor-docs/pull/80/files#r1506091213 (lines 38-70)

...and directly contradicts these entries:
- https://github.com/MetaMask/contributor-docs/pull/80/files#r1506093304 (lines 253-285)
- https://github.com/MetaMask/contributor-docs/pull/80/files#r1506095370 (lines 412-439)

DDDDDanica:
Hey @MajorLift thanks for your review, when i was taking the ticket, i just collected the points in [this ticket](https://github.com/MetaMask/contributor-docs/issues/69) and didn't notice your work, will remove now

docs/typescript.md
```diff
@@ -1,14 +1,170 @@
...
+  // Ensure properties adhere to the Person interface
+  return { name, age, occupation };
+}
+
+// Usage:
+const person1 = createPerson({ name: 'Alice', age: 30 }); // { name: "Alice", age: 30, occupation: "Unknown" }
+const person2 = createPerson({ name: 'Bob', age: 25, occupation: 'Developer' }); // { name: "Bob", age: 25, occupation: "Developer" }
+```
+
+TypeScript is planning to introduce the `satisfies` keyword, which will provide a more flexible way to declare that a data structure conforms to a specific type without sacrificing readability or maintainability. However, this feature is not yet available.
```
MajorLift, line 159: Type annotations prevent inference-based narrowing of the user-supplied type. This is why they should be avoided, or at least why type inference should generally be preferred (unless supplying the annotation narrows the type further than the inferred result).

In contrast, the `satisfies` operator enables the user to assign a wider/abstract type to a value as a constraint, while also allowing the value's type to be further narrowed.

It lets us avoid wrapping specific types in a function to assign a wider type.

> Need to create a data structure with specific values, but ensure that it's declared with a more abstract type? Create a function that takes an argument of the type you want, then use it to build the object. (https://github.com/MetaMask/contributor-docs/issues/69#issuecomment-1876114356)

```ts
const limitExceeded = () =>
    ({
      jsonrpc: '2.0',
      id: IdGenerator(),
      error: {
        code: -32005,
        message: 'Limit exceeded',
      },
    }),
```

I'll update this section with my PR for https://github.com/MetaMask/contributor-docs/issues/47, since I already cover these points there.

docs/typescript.md
```diff
@@ -1,14 +1,170 @@
...
 
-These guidelines specifically apply to TypeScript.
+**_Purpose_**: Establish best practices for writing robust, maintainable, and efficient TypeScript code.
 
-## Provide explicit return types for functions/methods
+**_Audience_**: Internal and external developers using TypeScript in Metamask.
+
+## **Interfaces**
+
+By defining expected object structures with interfaces, TypeScript promotes developer intent and unlocks powerful IDE features like accurate completion and context-aware navigation.
```
MajorLift, line 9: FYI: We're enforcing usage of type aliases over interfaces in all of our packages with the eslint rule `@typescript-eslint/consistent-type-definitions`.

See the following for previous discussions on this matter:
- https://github.com/MetaMask/eslint-config/pull/216
- https://github.com/MetaMask/core/pull/1933

docs/typescript.md
```diff
@@ -1,14 +1,170 @@
...
+  Left,
+  Right,
+}
+
+const directions: Direction[] = [Direction.Up, Direction.Down];
+
+for (const key in directions) {
+  const direction = directions[key] as Direction; // Assert to access enum values.
+  console.log(direction); // Output: Direction.Up, Direction.Down
+}
```
MajorLift, line 125: The type assertion can be easily avoided here:

```suggestion
for (const key of directions) {
  const direction = directions[key];
  console.log(direction); // Output: Direction.Up, Direction.Down
}
```

docs/typescript.md
```diff
@@ -1,14 +1,170 @@
...
+// Omit color if quantity is 1
+const singleItemPayload = Omit<CartItem, "color" extends string ? "color" : never>;
+
+// Omit color for all items if quantity is always 1
+const cartPayload: singleItemPayload[] = [];
+```
+
+#### Avoid type assertions, or explain their purpose whenever necessary.
+
+Type assertions inform TypeScript about a variable's actual type, using the `as` syntax or the angle bracket `<Type>` syntax.
```
MajorLift, line 80: Generic types are completely different from type assertions, especially since TypeScript doesn't do type casting.

See https://github.com/Microsoft/TypeScript/wiki/FAQ#why-doesnt-this-incorrect-cast-throw-a-runtime-error

```suggestion
Type assertions inform TypeScript about a variable's actual type, using the `as` syntax.
```

docs/typescript.md
```diff
@@ -1,14 +1,170 @@
...
+// before
+function createExampleMiddleware(
+  exampleParam,
+): JsonRpcMiddleware<JsonRpcParams, Json>;
+// after
+function createExampleMiddleware<
+  Params extends JsonRpcParams = JsonRpcParmas,
+  Result extends Json = Json,
+>(exampleParam): JsonRpcMiddleware<Params, Result>;
+```
```
MajorLift, line 39: This would be a much better example of generic constraints with the following change:

```suggestion
// before
function createExampleMiddleware<Params, Result>(
  exampleParam,
): JsonRpcMiddleware<JsonRpcParams, Json>;
// after
function createExampleMiddleware<
  Params extends JsonRpcParams,
  Result extends Json,
>(exampleParam): JsonRpcMiddleware<Params, Result>;
```
(removed the default params to keep the example focused on generic constraints and avoid confusion)

See: https://www.typescriptlang.org/docs/handbook/2/generics.html#generic-constraints

docs/typescript.md
```diff
@@ -1,14 +1,170 @@
...
-## Provide explicit return types for functions/methods
+**_Audience_**: Internal and external developers using TypeScript in Metamask.
+
+## **Interfaces**
+
+By defining expected object structures with interfaces, TypeScript promotes developer intent and unlocks powerful IDE features like accurate completion and context-aware navigation.
+
+### **Enumerations**
+
+TypeScript offers several tools for crafting clear data definitions, with enumerations and unions standing as popular choices.
```
MajorLift, line 13: The name for enumerated constants in TypeScript is `enum`s not "enumerations." 

```suggestion
### Enums

TypeScript offers several tools for crafting clear data definitions, with enums and unions standing as popular choices.
```

docs/typescript.md
```diff
@@ -1,14 +1,170 @@
-# TypeScript Guidelines
+# **Typescript guidelines**
```
MajorLift, line 1:
```suggestion
# Typescript Guidelines
```
- All words should be capitalized in document title. 
- Applying bold style to a heading doesn't have any effect. All instances of this should be fixed throughout the doc.

MajorLift (commented):
@DDDDDanica No worries, thanks for responding quickly! I had a few more suggestions and corrections:

docs/typescript.md
MajorLift: Also, `for .. in` loops are banned by our eslint rules.
- For previous discussion, see https://github.com/MetaMask/core/pull/1865

docs/typescript.md
MajorLift: Looks like this is a deprecated syntax for type assertions. In any case, it should be removed as it's no longer in use.

docs/typescript.md
DDDDDanica: Gotcha !

docs/typescript.md
DDDDDanica: Will change the title to `type`, thanks for the heads up

docs/typescript.md
DDDDDanica: yes added !

docs/typescript.md
DDDDDanica: yes, added !

docs/typescript.md
DDDDDanica: Thanks for the heads up, applied the changes now. thanks so much !

docs/typescript.md
```diff
@@ -1,14 +1,209 @@
...
 
-These guidelines specifically apply to TypeScript.
+**_Purpose_**: Establish best practices for writing robust, maintainable, and efficient TypeScript code.
 
-## Provide explicit return types for functions/methods
+**_Audience_**: Internal and external developers using TypeScript in Metamask.
+
+## **Type**
+
+By defining expected object structures with interfaces, TypeScript promotes developer intent and unlocks powerful IDE features like accurate completion and context-aware navigation.
```
MajorLift, line 9: (continued from https://github.com/MetaMask/contributor-docs/pull/80#discussion_r1506360915)

On second glance, we do use interfaces in the sense that you're describing it. And I think establishing contracts for classes is a development pattern worth encouraging. We just consistently use type alias syntax instead of interface syntax to write all types, including class "interfaces." Sorry for the confusion here. 

How about we split this into two sections?

```suggestion
## `implements` keyword

The `implements` keyword enables us to define and enforce interfaces, i.e. strict contracts consisting of expected object and class properties and abstract method signatures. 

Writing an interface to establish the specifications of a class that external code can interact with without being aware of internal implementation details is encouraged as sound OOP development practice. 

Here's an abbreviated example from `@metamask/polling-controller` of an interface being used to define one of our most important constructs.

\```ts
export type IPollingController = {
  ...
}
...
export function AbstractPollingControllerBaseMixin<TBase extends Constructor>(
  Base: TBase,
) {
  abstract class AbstractPollingControllerBase
    extends Base
    implements IPollingController
  { ... }
  return AbstractPollingControllerBase  
}
\```

The concept of the interface as discussed in this section is not to be confused with interface syntax as opposed to type alias. Note that in the above example, the `IPollingController` interface is defined as a type alias, not using the `interface` keyword. 

## Always prefer Type Aliases over Interface syntax

We enforce consistent and exclusive usage of type aliases over interface syntax for several reasons:

- The capabilities of type aliases is a strict superset of those of interfaces. 
  - Crucially, `extends`, `implements` are also supported by type aliases.
  - Declaration merging is the only exception, but we have no use case for this feature that cannot be substituted by using type intersections.
- Unlike interfaces, type aliases extend `Record` and have an index signature of `string` by default, which makes them compatible with our Json-serializable types (most notably `Record<string, Json>`). 
- Interfaces do not support multiple inheritance, while type aliases can be freely merged using the intersection (`&`) operator.
```

(remove the backticks around the fenced code block)

This is intended as an initial draft. Feel free to edit.

docs/typescript.md
```diff
@@ -1,14 +1,209 @@
...
+  Right,
+}
+
+const directions: Direction[] = [Direction.Up, Direction.Down];
+
+for (const key of directions) {
+  const direction = directions[key];
+  console.log(direction); // Output: Direction.Up, Direction.Down
+}
+```
```
MajorLift, line 165: This is no longer an example of `as` assertion being used. Should we replace it with one that fits under the section heading?

> #### Avoid type assertions, or explain their purpose whenever necessary.

docs/typescript.md
```diff
@@ -1,14 +1,209 @@
...
+
+## **Type**
+
+By defining expected object structures with interfaces, TypeScript promotes developer intent and unlocks powerful IDE features like accurate completion and context-aware navigation.
+
+### **Enums**
+
+TypeScript offers several tools for crafting clear data definitions, with enumerations and unions standing as popular choices.
+
+**Consider using enums over union types for situations with a fixed set of known values.**
```
MajorLift, line 15: The entries in this doc should be at the same heading level.

```suggestion
#### Consider using enums over union types for situations with a fixed set of known values.
```

docs/typescript.md
```diff
@@ -1,14 +1,209 @@
...
+
+```typescript
+enum AccountType {
+  Admin = 'admin',
+  User = 'user',
+  Guest = 'guest',
+}
+```
+
+**Don't use numeric enums**
```
MajorLift, line 35: The entries in this doc should be at the same heading level.

```suggestion
#### Don't use numeric enums
```

docs/typescript.md
```diff
@@ -1,14 +1,209 @@
...
+**_Purpose_**: Establish best practices for writing robust, maintainable, and efficient TypeScript code.
 
-## Provide explicit return types for functions/methods
+**_Audience_**: Internal and external developers using TypeScript in Metamask.
+
+## **Type**
+
+By defining expected object structures with interfaces, TypeScript promotes developer intent and unlocks powerful IDE features like accurate completion and context-aware navigation.
+
+### **Enums**
```
MajorLift, line 11: Applying bold style to a markdown heading has no effect.

```suggestion
### Enums
```

docs/typescript.md
```diff
@@ -1,14 +1,209 @@
...
+  Up = 'Up',
+  Down = 'Down',
+  Left = 'Left',
+  Right = 'Right',
+}
+
+const directions = Object.values(Direction); // ["Up", "Down", "Left", "Right"]
+```
+
+## **Types**
```
MajorLift, line 64: Applying bold style to a markdown heading has no effect.

```suggestion
## Types
```

docs/typescript.md
```diff
@@ -1,14 +1,209 @@
-# TypeScript Guidelines
+# Typescript Guidelines
 
-These guidelines specifically apply to TypeScript.
+**_Purpose_**: Establish best practices for writing robust, maintainable, and efficient TypeScript code.
 
-## Provide explicit return types for functions/methods
+**_Audience_**: Internal and external developers using TypeScript in Metamask.
```
MajorLift, line 5:
```suggestion
**_Audience_**: Internal and external developers using TypeScript to contribute to MetaMask.
```

We want to use consistent styling for our main product name (similarly, "Consensys," not "ConsenSys" or some other variation).

MajorLift (commented):
Thanks for applying the suggestions! Some more notes:

docs/typescript.md
MajorLift: Continued in https://github.com/MetaMask/contributor-docs/pull/80#discussion_r1506955448

docs/typescript.md
MajorLift: Continued in https://github.com/MetaMask/contributor-docs/pull/80#discussion_r1506956360

docs/typescript.md
MajorLift: Removing the second entry is also acceptable since `@typescript-eslint/consistent-type-definitions` does a good job of enforcing the practice without everyone having to know the reasoning behind its adoption. 

In general, the guidelines doesn't need to cover conventions that we already enforce using linter rules or static code analysis in too much detail. This doc isn't intended as a comprehensive collection of every preference we have agreed upon. The priority should be on covering principles or processes that aren't reducible to simple rulesets and rely on a deeper understanding from the developers to be effective.

docs/typescript.md
MajorLift: Ah actually this is my bad. I see this example is working by coincidence: because `Direction.Up` and `Direction.Down` happen to be the first two index keys of the `directions` array.

This would still work as an example of type assertion with the following changes:

```typescript
enum Direction {
  Up = 'up',
  Down = 'down',
  Left = 'left',
  Right = 'right',
}

const directions: Direction[] = [Direction.Up, Direction.Down];

for (const key of directions) {
  const direction = directions[key as keyof typeof directions];
  console.log(direction); // Output: Direction.Up, Direction.Down
}
```

We also have a built-in utility type for narrowing key signatures, which simplifies the code.

```ts
import { getKnownPropertyNames } from '@metamask/utils';

...

for (const key of getKnonwPropertyNames(direction)) {
  const direction = directions[key];
}
```

docs/typescript.md
DDDDDanica: Thanks for all the recommendations, i've added your suggestions, and changed the section title as `Implements`

docs/typescript.md
DDDDDanica: I'm using `**` to represent lower level title, but can certainly change everyone !

docs/typescript.md
DDDDDanica: changed also

docs/typescript.md
DDDDDanica: Thanks for the suggestion, i've learnt a new`getKnonwPropertyNames` today

docs/typescript.md
```diff
@@ -149,18 +179,18 @@ console.log(youngestUser); // { name: "...", age: maxAge }
...
+enum Direction { 
+    Up = 'up', 
+    Down = 'down', 
+    Left = 'left', 
+    Right = 'right', 
+} 
+const directions: Direction[] = [Direction.Up, Direction.Down]; 
+for (const key of getKnonwPropertyNames(direction)) {
+    const direction = directions[key as keyof typeof directions];
+    console.log(direction); // Output: Direction.Up, Direction.Down }
```
MajorLift, line 193: I should have clarified that my `getKnownPropertyNames` example wasn't meant to be used here. I'm sorry for the confusion.

The usefulness of `getKnownPropertyNames` is that it makes the `as` assertion *unnecessary* ([you can see why by checking out its definition](https://github.com/MetaMask/utils/blob/ec0237cf6793b9b31efffde247152c12ffd1a260/src/misc.ts#L102-L117)). So the above code can't be used as an example for this section.

We'll need to remove the `getKnownPropertyNames` import statement. The type annotation for `directions` can also be removed (this is preferable for reasons detailed [here](https://github.com/MetaMask/contributor-docs/pull/74/files#diff-ae03a9f2eb638a25c1d8d7ba853c78220f1ed44a19ad97a8d780cbc0c00a4260R7-R46)).

```ts
enum Direction { 
  Up = 'up', 
  Down = 'down', 
  Left = 'left', 
  Right = 'right', 
}
const directions = [Direction.Up, Direction.Down];
```

From there, we can resolve the index signature error that arises from the property access expression `directions[key]`,
> Element implicitly has an 'any' type because index expression is not of type 'number'.(7015)

by choosing one of these two options:

- option 1:
```ts
for (const key of Object.keys(direction)) {
  const direction = directions[key as keyof typeof directions];
}
```
- option 2:
```ts
for (const key of Object.keys(directions) as (keyof typeof directions)[]) {
  const direction = directions[key];
}
```

You can use either or both as the example(s) for this entry.

docs/typescript.md
```diff
@@ -2,17 +2,47 @@
...
 **_Purpose_**: Establish best practices for writing robust, maintainable, and efficient TypeScript code.
 
-**_Audience_**: Internal and external developers using TypeScript in Metamask.
+**_Audience_**: Internal and external developers using TypeScript to contribute to MetaMask.
 
-## **Type**
+### Implements
 
-By defining expected object structures with interfaces, TypeScript promotes developer intent and unlocks powerful IDE features like accurate completion and context-aware navigation.
+**`implements` keyword**
```
MajorLift, line 9: Nit: is this maybe a section heading rather than just emphasized text?

docs/typescript.md
```diff
@@ -2,17 +2,47 @@
...
+        extends Base
+        implements IPollingController
+    { ... }
+    return AbstractPollingControllerBase
+}
+```
+
+The concept of the interface as discussed in this section is not to be confused with interface syntax as opposed to type alias syntax. Note that in the above example, the `IPollingController` interface is defined as a type alias, not using the `interface` keyword.
+
+**Always prefer type aliases over `interface` keyword**
```
MajorLift, line 32: Nit: is this maybe a section heading rather than just emphasized text?

docs/typescript.md
```diff
@@ -2,17 +2,47 @@
...
+    - Declaration merging is the only exception, but we have no use case for this feature that cannot be substituted by using type intersections.
+- Unlike interfaces, type aliases extend `Record` and have an index signature of `string` by default, which makes them compatible with our Json-serializable types (most notably `Record<string, Json>`).
+- Interfaces do not support multiple inheritance, while type aliases can be freely merged using the intersection (`&`) operator.
+
+### Enums
 
 TypeScript offers several tools for crafting clear data definitions, with enumerations and unions standing as popular choices.
 
-**Consider using enums over union types for situations with a fixed set of known values.**
+####Consider using enums over union types for situations with a fixed set of known values.
```
MajorLift, line 45: Nit:
> MD018/no-missing-space-atx: No space after hash on atx style heading markdownlint [MD018](https://github.com/DavidAnson/markdownlint/blob/v0.33.0/doc/md018.md)

```suggestion
#### Consider using enums over union types for situations with a fixed set of known values.
```

docs/typescript.md
```diff
@@ -32,7 +62,7 @@ enum AccountType {
 }
 ```
 
-**Don't use numeric enums**
+####Don't use numeric enums
```
MajorLift, line 65: Nit:
> MD018/no-missing-space-atx: No space after hash on atx style heading markdownlint [MD018](https://github.com/DavidAnson/markdownlint/blob/v0.33.0/doc/md018.md)

```suggestion
#### Don't use numeric enums
```

MajorLift (commented):
Looks like we're almost there! Just one important fix and some linting issues left.

On the subject of linting, I recommend setting up the [`markdownlint` extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) in your IDE. Its ruleset is mostly consistent with our prettier config and a lot more comprehensive.

Having a shared set of conventions on how to structure and format these docs helps future collaborators add their contributions without feeling like they're stepping on any toes.

On why consistency in the heading hierarchy is a good thing, check out the outline view in the preview for this doc: https://github.com/MetaMask/contributor-docs/blob/feature/69-typescript/docs/typescript.md. You can enable it by clicking on the unordered-list icon on the top right (`Button.types__StyledButton-sc-ws60qy-0`).

I think we should be ready to merge after this next round of fixes. Thanks for bearing with me through all these comments.

docs/typescript.md
DDDDDanica: Good catch !! changed !

docs/typescript.md
DDDDDanica: Thank you @MajorLift for the comprehensive explanation, modified this section as well

MajorLift (dismissed):
LGTM! 

Added a commit https://github.com/MetaMask/contributor-docs/pull/80/commits/5d990cd909b31ac6dfa3380ec73b9201a8c261dc before leaving the approval so you could merge without more back and forth. 

Feel free to ping me for another approval if there are more changes you want to add.

MajorLift (approved):

DDDDDanica:
@MajorLift thanks for your review and approval, they are really helpful ! I'll draft another PR to address the rest

DDDDDanica:
Have a conversation with @MajorLift offline to merge his work first, then add more for <TODO> that he will leave for me

bdstar81 (approved):

PR #83, opened 2024-03-13
feat: Expand advice about small pull requests

Gudahtt:
The section about creating small pull requests has been updated to include more details and advice.


mcmire (approved):
Makes sense!

PR #84, opened 2024-03-13
feat: Add example about validating objects from external sources

Gudahtt:
An example has been added to our "Data Validation and Sanitization" guideline about the risk of dynamic behavior on objects from external sources.

This was originally suggested here: https://github.com/MetaMask/contributor-docs/pull/58#discussion_r1379781336 The suggest has been reformatted/reworded here in this PR to match the context.


mcmire (approved):
Makes sense!

PR #86, opened 2024-03-13
feat: Add more vulnerability scanning tool examples

Gudahtt:
These examples were suggested in this comment: https://github.com/MetaMask/contributor-docs/pull/58#discussion_r1379799956

I was going to include further advice on how to use Dependabot, but I am not yet sure what advice to offer here because we don't yet have a Dependabot workflow that we're happy with. It's currently disabled in some places, and it's configured ineffectively in others. We can revisit this later.


mcmire (approved):
LGTM.

PR #93, opened 2024-05-06
Fix 'Highlight exercise phase' example

mcmire:
The `describe` block was incorrect, and the examples needed a bit more explanation.


Gudahtt (approved):
LGTM!

PR #103, opened 2024-08-06
doc: Add Redux Performance Guidelines

MarioAslau:
Adding guidelines on how to improve performance for Redux and also Best practices


gambinish:
Do we have any sort of caching strategy on the mobile app? In extension, we have `fetchWithCache` which caches the result of an API request so that it isn't over fetched.

Useful with things like fetching `safeChains` or something along those lines.

docs/redux.md
```diff
@@ -0,0 +1,1228 @@
...
+            allIds: [1, 2],
+        },
+        posts: {
+            byId: {
+                1: { id: 1, title: 'Post 1' },
+                2: { id: 2, title: 'Post 2' },
+            },
+            allIds: [1, 2],
+        },
+    };
```
NicolasMassart, line 113: C: it may not be obvious that these are IDs as they look like indexes. If it's just indexes we could use an array instead, right?
So maybe replace 1, 2 by things that look more like ids like short random hex string for the example to be clear?

docs/redux.md
```diff
@@ -0,0 +1,1228 @@
...
+        };
+        case 'SET_EXPENSIVE_RESULT':
+        return {
+            ...state,
+            expensiveResult: action.payload,
+        };
+        default:
+            return state;
+    }
+}
```
NicolasMassart, line 48: C: i like the good/bad examples.
Maybe make that more explicit and visible with this outside of the code block and add emoji?
Suggestion:

🚫 Bad: Performing expensive calculations inside the reducer:
```typescript
const initialState = { data: [], expensiveResult: 0 };
// the bad code example
...
}
```
👍 Good: Perform expensive calculations outside the reducer:
```typescript
const initialState = { data: [], expensiveResult: 0 };
// the good code example
...
}
```

docs/redux.md
```diff
@@ -0,0 +1,1228 @@
...
+
+// Good: Storing serializable values in state
+const initialState = { data: {} };
+
+function myReducer(state = initialState, action) {
+  switch (action.type) {
+    case 'SET_DATA':
+      return {
+        ...state,
+        data: action.payload, // payload is a plain object
```
NicolasMassart, line 300: Q: should we convert to JSON string first? Is there a preferred format?

docs/redux.md
```diff
@@ -0,0 +1,1228 @@
...
+      return state;
+  }
+}
+```
+
+## 3.1.4 **Only One Redux Store Per App**
+
+Reducers should only depend on their state and action arguments.
+
+```typescript
```
NicolasMassart, line 312:
```suggestion
```tsx
```

docs/redux.md
```diff
@@ -0,0 +1,1228 @@
...
+  },
+});
+
+export const { increment, decrement } = counterSlice.actions;
+export default store;
+```
+
+## 3.2.2 **Use Immer for Writing Immutable Updates**
+
+Immer allows you to write simpler immutable updates using "mutative" logic.
```
NicolasMassart, line 365: S: add links to external library docs
```suggestion
[Immer](https://redux-toolkit.js.org/usage/immer-reducers) allows you to write simpler immutable updates using "mutative" logic.
```

docs/redux.md
```diff
@@ -0,0 +1,1228 @@
...
+      return state;
+  }
+};
+```
+
+## 3.2.3 **Structure Files as Feature Folders with Single-File Logic**
+
+Co-locating logic for a given feature in one place typically makes it easier to maintain that code.
+
+```typescript
```
NicolasMassart, line 388:
```suggestion
```text
```

docs/redux.md
```diff
@@ -0,0 +1,1228 @@
...
+      return state;
+    default:
+      return state;
+  }
+}
+```
+
+## 3.2.9 **Normalize Complex Nested/Relational State**
+
+Prefer storing data in a "normalized" form.
```
NicolasMassart, line 525: C: si this similar to "2.3 Normalization" section?

docs/redux.md
```diff
@@ -0,0 +1,1228 @@
...
+    payload: user,
+  };
+}
+```
+
+## 3.2.15 **Evaluate Where Each Piece of State Should Live**
+
+Decide what state should live in the Redux store and what should stay in component state.
+
+```typescript
```
NicolasMassart, line 648:
```suggestion
```tsx
```

docs/redux.md
```diff
@@ -0,0 +1,1228 @@
...
+    </form>
+  );
+}
+```
+
+## 3.2.16 **Use the React-Redux Hooks API**
+
+Prefer using the React-Redux hooks API.
+
+```typescript
```
NicolasMassart, line 673:
```suggestion
```tsx
```

docs/redux.md
```diff
@@ -0,0 +1,1228 @@
...
+    </div>
+  );
+}
+```
+
+## 3.2.17 **Connect More Components to Read Data from the Store**
+
+Prefer having more UI components subscribed to the Redux store and reading data at a more granular level.
+
+```typescript
```
NicolasMassart, line 693:
```suggestion
```tsx
```

NicolasMassart (commented):
So much good stuff in here! Thanks for this guide!

docs/redux.md
MarioAslau: Converting data to a JSON string is one way to ensure data is serializable, but it is not always necessary or preferred. Instead, we should aim to use plain JavaScript objects and arrays

docs/redux.md
```diff
@@ -0,0 +1,1228 @@
...
+    const state = getState();
+    // Perform complex logic here
+    dispatch({ type: 'COMPLEX_LOGIC_DONE' });
+  };
+};
+```
+
+## 3.3.7 **Use Selector Functions to Read from Store State**
+
+Use memoized selector functions for reading store state whenever possible.
```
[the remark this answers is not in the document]
MarioAslau: Same as above

NicolasMassart (approved):
Looks good to me.

MajorLift (approved):
LGTM! Impressive scope :)

docs/redux.md
```diff
@@ -0,0 +1,1253 @@
...
+  <Provider store={store}>
+    <App />
+  </Provider>,
+  document.getElementById('root'),
+);
+```
+
+## 3.2 **Priority B Rules: Strongly Recommended**
+
+## 3.2.1 **Use Redux Toolkit for Writing Redux Logice**
```
matthewwalsh0, line 359: Minor typo on `Logice`.

docs/redux.md
```diff
@@ -0,0 +1,1253 @@
...
+## 3.3.7 **Use Selector Functions to Read from Store State**
+
+Use memoized selector functions for reading store state whenever possible.
+
+```tsx
+import { createSelector } from 'reselect';
+
+const selectTodos = (state) => state.todos;
+
+const selectVisibleTodos = createSelector(
```
matthewwalsh0, line 962: It's hopefully a temporary problem while we improve the state architecture, but should we mention `createDeepEqualSelector`?

MarioAslau: Based on a performance talk we had last week looking at Sentry reports, @Gudahtt mentioned that the work they are doing on controllers (Updated Controllers until all controllers extend Base controller V2.) they will use Immer under the hood (same as RTK), enabling us to only use shallow comparison which will be more performant than our deep comparison selectors.

matthewwalsh0: A current limitation in the extension is that the background sends all state to the frontend over a JSON-RPC API so it can't distinguish shallow reference changes.

But I'm actively working on a PR to resolve this, so happy for it not to be included here if it's been considered not best practice but a temporary workaround.

PR #105, opened 2024-08-30
Update e2e-testing.md

chrisleewilcox:
Discussed the use of the term 'should' at the start of test names.  Using 'should' at the start of test names is a long standing and accepted standard.  Our guidelines should be more aligned with the tools and frameworks guidelines and recommendations that we use.

Even repeated test names within a test file using 'should' at the start of test names should be considered valid.

The test framework we use implies all test names start with 'it' which makes 'should' coming after 'it' a more readable, expressive, and consistent practice.

### DIFF
--- a/docs/e2e-testing.md
+++ b/docs/e2e-testing.md
@@ -30,14 +30,7 @@ The test name should communicate the purpose and behaviour of the test. A clear 
 - send 1 TST to Bob
 ```
 
-⚠️ Test names remain approachable, recommend to avoid using the meaningless prefix 'should'
-
-```javascript
-- should add Bob to the address book
-- should send 1 TST to Bob
-```
-
-❌ Test name should be completely avoided: The use of a `should` prefix and the word `and` can decrease the readability of the test,making it harder to understand what the test is doing as well as diagnose and fix issues.
+❌ Test name should be completely avoided: Long test names that may over explain the test intention and the word `and` can decrease the readability of the test,making it harder to understand what the test is doing as well as diagnose and fix issues.
 
 ```javascript
 - should add Bob to the address book and send 1 TST to Bob
--- a/docs/e2e-testing.md
+++ b/docs/e2e-testing.md
@@ -30,7 +30,7 @@ The test name should communicate the purpose and behaviour of the test. A clear 
 - send 1 TST to Bob
 ```
 
-❌ Test name should be completely avoided: Long test names that may over explain the test intention and the word `and` can decrease the readability of the test,making it harder to understand what the test is doing as well as diagnose and fix issues.
+❌ Test name should be completely avoided: Long test names that may over explain the test intention and the word `and` can decrease the readability of the test, making it harder to understand what the test is doing as well as diagnose and fix issues.
 
 ```javascript
 - should add Bob to the address book and send 1 TST to Bob


hjetpoluru (approved):
LGTM

docs/e2e-testing.md
```diff
@@ -30,14 +30,7 @@ The test name should communicate the purpose and behaviour of the test. A clear
...
 
-⚠️ Test names remain approachable, recommend to avoid using the meaningless prefix 'should'
-
-```javascript
-- should add Bob to the address book
-- should send 1 TST to Bob
-```
-
-❌ Test name should be completely avoided: The use of a `should` prefix and the word `and` can decrease the readability of the test,making it harder to understand what the test is doing as well as diagnose and fix issues.
+❌ Test name should be completely avoided: Long test names that may over explain the test intention and the word `and` can decrease the readability of the test,making it harder to understand what the test is doing as well as diagnose and fix issues.
```
desi, line 33: I believe the reason this was written this way originally was because this was a pattern we were seeing and wanted to call out specifically that we don't want to continue using that pattern. Can you help me understand why you want to remove that language?

Gudahtt:
This was a deliberate choice to make test descriptions more concise. By re-framing the description to say what a test _does_, rather than what it _should do_, you convey the same information with fewer words.

e.g.
```
it('should show a warning when a malicious transaction is suggested')
```

versus

```
it('shows a warning when a malicious transaction is suggested')
```

docs/e2e-testing.md
Gudahtt, line 33: This new explanation for why the below test name should be avoided seems inaccurate. It's not a particularly long description, and it does not over explain the test intention.

There is also a missing space:
```suggestion
❌ Test name should be completely avoided: Long test names that may over explain the test intention and the word `and` can decrease the readability of the test, making it harder to understand what the test is doing as well as diagnose and fix issues.
```

chrisleewilcox: The pattern of using 'should' at the beginning of a test description is to make the test worded in a manner that if spoken is sounds right and makes sense.  It is a very common QA industry practice when using BDD test framework for test development.  This practice is not a preference but more of a mechanism, or method, to easily name test without having to think too much about it.  When the methods becomes a habit then QA engineers would just always invoke that habit even when it may not be needed in all or even most cases.  Think of it like 'it should' at the beginning is compatible with all test naming and that's why it's widely used.

chrisleewilcox: The edit was to adjust from saying never to use 'should' and to emphasize clear what the test is intended.  But never use 'should' may be a little extreme to be applied in all cases as there are clear examples of using 'should' just sounds the most appropriate for a test.

We can adjust to say something like....

`Although 'should' is a widely used method for naming test it does not mean it needs to be always used and it's encouraged to name test in any manner that makes the test understandable.  Some teams may have a preference for naming conventions so check with teams on desired approach.`

chrisleewilcox: Here is a prompt that may explain better the use of a common practice for naming test.

`For BDD test development, why is 'should' commonly used at the start of it()?`
image

desi: @chrisleewilcox I come from old school Ruby world / rspec so I am very familiar with the BDD. My main concern here was that I knew this call out was very deliberate and I didn't want it to change without some additional discussion. I believe @mcmire was one of the folks that felt strongly about this and he is out on paternity leave. He may have additional thoughts when he returns but for now I do really like the rewording you included in the comment above. I think the new wording is a nice balance, essentially "don't use it just because you are in the habit of it but feel free to use it if it actually makes the test easier to read"

@Gudahtt do you have additional thoughts?

chrisleewilcox: @desi thanks for the feedback.  Spoke with some QA team members and some of the responses were related to test output that didn't make sense outside of QA engineers.  Now that I have a better understanding of the history and reasoning I have no problem leaving the doc as is. 

I'll just add @Gudahtt recommended change for proper spacing.

PR #107, opened 2024-09-05
Add snapshot guidelines

NicolasMassart:
Add guidelines on how to use and name snapshot unit tests


joaoloureirop (approved):

tommasini (approved):

docs/unit-testing.md
```diff
@@ -1070,3 +1070,35 @@ Jest incorporates most of the features of Sinon with a slimmer API:
 - `jest.fn()` can be used in place of `sinon.stub()`.
 - `jest.spyOn(object, method)` can be used in place of `sinon.spy(object, method)` or `sinon.stub(object, method)` (with the caveat that the method being spied upon will still be called by default).
 - `jest.useFakeTimers()` can be used in place of `sinon.useFakeTimers()` (though note that Jest's "clock" object had fewer features than Sinon's prior to Jest v29.5).
+
+## Snapshots
+
+Jest snapshots are not testing the validity of the value tested against a snapshot.
```
Gudahtt, line 1076: Hmm, inline snapshots may be an exception to this? Assuming that they're readable inline snapshots (which they should be, that's the whole point of them being inline).

But maybe that's obvious.

Gudahtt (approved):
LGTM!

docs/unit-testing.md
NicolasMassart: Even inline, they only test the value hasn't changed, but not that the value is right.

PR #110, opened 2024-10-21
chore: update for react hooks guidelines

Daniel-Cross:
Updating the user guidelines for React hooks.

### DIFF
--- a/docs/react.md
+++ b/docs/react.md
@@ -2,8 +2,115 @@
 
 ## Performance
 
-Note that the below are purely optimizations so any code should still function without them.
+TL;DR:
+- `useMemo` and `useCallback` should only be used when you are having legitimate performance issues. They should not be used as a precaution or as a pre-optimisation. Both hooks allocate device memory to watch for changes. The more hooks you have, the more memory is allocated. The longer the app is open, the more memory is allocated. This is a self inflicted memory leak. Please use for long lists of data that are expensive to calculate and do not need re-renders, but not for simple objects or primitives.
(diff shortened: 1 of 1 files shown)


sethkfman (approved):
@Daniel-Cross Thanks for getting these added. These are great.

dabossdana (approved):

MajorLift:
Perhaps this is out of scope for this PR, but so long as we're discussing memoization tactics specifically, should we add some general remarks about React Compiler? This seems especially relevant now with the [new react compiler beta release](https://react.dev/blog/2024/10/21/react-compiler-beta-release) that supports react v17-18.

It appears that in general manual memoization will be stripped and replaced by React Compiler, and therefore could be removed altogether eventually. It doesn't seem like this should affect our current code.

However, an exception is memoization that is intended to ensure correctness instead of performance benefits. Maybe we should add a guideline to highlight any new additions that fall under this category with a comment, so that we know not to remove them in the future.

> It is generally safe to leave manual memoization in your code – the compiler understands it but will compile in its own lower-level form of memoization even for manually memoized code. The compiler does however check its memoization against manual memoization to ensure that we memoized it exactly the same. The reason we need to do this is in case the manual memoization was used for correctness. For example, if a manually memoized value was passed as a dependency to a useEffect call somewhere else and it wasn't memoized in the same way, this might cause over/under firing of effects or even infinite loops.
> 
> For this reason, our current recommendation is to leave manual memoization in for existing code, but for new code written after the compiler is added you can omit them.
>
> -- https://github.com/reactwg/react-compiler/discussions/16#discussioncomment-9991176

PR #112, opened 2024-11-07
feat: Add testing principles

Gudahtt:
A section has been added to the "Testing Overview" document to describe our testing principles. These establish what the goals of our automated test suite are, and help contributors differentiate between good and bad tests.

[Rendered](https://github.com/MetaMask/contributor-docs/blob/document-testing-purpose/docs/testing/overview.md)

### DIFF
--- a/docs/testing/overview.md
+++ b/docs/testing/overview.md
@@ -6,13 +6,13 @@ Our automated tests fall into three general categories: end-to-end (e2e), integr
 
 ## Testing principles
 
-What makes a good test? Here are some general principles.
+The following are general principles about how to write automated tests effectively, applicable to all types of automated tests.
 
 ### 1. Tests should describe expectations
 
 What's the difference between a bug and a feature? A feature is expected, a bug is not.
 
-If it's not clear what the code-under-test is supposed to do, it won't be clear what to do when the test fails either. Clear expectations are essential for an effective test.
+Each test should be verifying specific expectations. If it's not clear what the code-under-test is supposed to do, it won't be clear what to do when the test fails either. Clear expectations are essential for an effective test.
 
 ### 2. Tests should catch bugs
 
@@ -44,9 +44,9 @@ What we consider "fast" depends on what the test does; what's fast for an end-to
 
 If a test doesn't align with the previous 6 principles, it's probably doing more harm than good. Consider deleting it.
 
-That doesn't mean we should skip writing tests with it's too difficult. On the contrary; we have high expectations for quality, we don't have the luxury of writing only the easiest most valuable tests. We need to aim for total coverage.
+That doesn't mean we should skip writing tests when it's too difficult. On the contrary, given our high expectations for quality we don't have the luxury of writing only the easiest tests. We need to aim for total coverage.
 
-But writing and maintaining bad tests is not the shortest path to complete test coverage. Our time is better spent writing good tests, and making investments that allow us to write more good tests. All code can be designed to be easy to test, and even the most complex test setup/teardown can be reasonably fast and easy to read/write with the right tools.
+However, writing and maintaining bad tests is not the shortest path to complete test coverage. Our time is better spent writing good tests, and making investments that allow us to write more good tests. All code can be designed to be easy to test, and even the most complex test setup/teardown can be reasonably fast and easy to read/write with the right tools.
 
 ## What are unit tests?
 
--- a/docs/testing/overview.md
+++ b/docs/testing/overview.md
@@ -1,6 +1,52 @@
 # Testing Overview
 
-At MetaMask, we are committed to delivering high-quality software. A key aspect of achieving this goal is through a robust testing strategy that includes both end-to-end (e2e), integration and unit tests. This document provides comprehensive guidelines on each testing type, emphasizing the importance of choosing the right test for the right scenario and detailing the specific practices for writing and organizing these tests.
+At MetaMask, we are committed to delivering high-quality software. These guidelines describe how we use automated tests to maximize product quality while minimizing the cost of writing and maintaining tests.
+
+Our automated tests fall into three general categories: end-to-end (e2e), integration and unit tests. This document provides comprehensive guidelines on each testing type, emphasizing the importance of choosing the right test for the right scenario and detailing the specific practices for writing and organizing these tests.
+
+## Testing principles
+
+What makes a good test? Here are some general principles.
+
+### 1. Tests should describe expectations
+
+What's the difference between a bug and a feature? A feature is expected, a bug is not.
+
+If it's not clear what the code-under-test is supposed to do, it won't be clear what to do when the test fails either. Clear expectations are essential for an effective test.
+
+### 2. Tests should catch bugs
+
+This is the main purpose of tests. Take care to ensure that each test actually fails when the described expectation is not met.
+
+### 3. Tests should not break often
+
+It's expected that tests will break when we change the behavior or APIs under test. But if a test breaks often from refactors, or from changes to behavior or APIs not directly under test, it's sign that the test may be poorly designed.
+
+Test breakages are expensive. We should do what we can to minimize them.
+
+### 4. Tests should be easy to read
+
+A test isn't effective if we can't understand why it failed. The more readable a test is, the faster we can fix it.
+
+### 5. Tests should be easy to write
+
+Some behaviors are inherently more complex than others, increasing the difficulty of testing them. But in general we want tests to be easy to write, so that we can keep costs low.
+
+If a test is hard to write, consider that as a warning that you might be approaching the problem in a non-optimal way. Resist the urge to power through the work, and instead consider how redesigning the code-under-test or using a different testing strategy might save time.
+
+### 6. Tests should be fast
+
+Making tests faster has a dramatic improvement on productivity, reducing feedback loops.
+
+What we consider "fast" depends on what the test does; what's fast for an end-to-end test might be slow for a unit test. But faster is always better, for all types of tests.
+
+### 7. Tests should have a high return on investment
+
+If a test doesn't align with the previous 6 principles, it's probably doing more harm than good. Consider deleting it.
+
+That doesn't mean we should skip writing tests with it's too difficult. On the contrary; we have high expectations for quality, we don't have the luxury of writing only the easiest most valuable tests. We need to aim for total coverage.
+
+But writing and maintaining bad tests is not the shortest path to complete test coverage. Our time is better spent writing good tests, and making investments that allow us to write more good tests. All code can be designed to be easy to test, and even the most complex test setup/teardown can be reasonably fast and easy to read/write with the right tools.
 
 ## What are unit tests?
 
--- a/docs/testing/overview.md
+++ b/docs/testing/overview.md
@@ -20,7 +20,7 @@ This is the main purpose of tests. Take care to ensure that each test actually f
 
 ### 3. Tests should not break often
 
-It's expected that tests will break when we change the behavior or APIs under test. But if a test breaks often from refactors, or from changes to behavior or APIs not directly under test, it's sign that the test may be poorly designed.
+It's expected that tests will break when we change the behavior or APIs under test. But if a test breaks often from refactors, or from changes to behavior or APIs not directly under test, it's sign that the test (or the code-under-test) may be poorly designed.
 
 Test breakages are expensive. We should do what we can to minimize them.
 


docs/testing/overview.md
```diff
@@ -1,6 +1,52 @@
 # Testing Overview
 
-At MetaMask, we are committed to delivering high-quality software. A key aspect of achieving this goal is through a robust testing strategy that includes both end-to-end (e2e), integration and unit tests. This document provides comprehensive guidelines on each testing type, emphasizing the importance of choosing the right test for the right scenario and detailing the specific practices for writing and organizing these tests.
+At MetaMask, we are committed to delivering high-quality software. These guidelines describe how we use automated tests to maximize product quality while minimizing the cost of writing and maintaining tests.
+
+Our automated tests fall into three general categories: end-to-end (e2e), integration and unit tests. This document provides comprehensive guidelines on each testing type, emphasizing the importance of choosing the right test for the right scenario and detailing the specific practices for writing and organizing these tests.
+
+## Testing principles
```
Gudahtt, line 7: Maybe I should move this to a separate document, and link it here?

Gudahtt:
Inspiration for these comes from:
* Some input from @cryptotavares 
* This talk: https://www.youtube.com/watch?v=_pnW-JjmyXE (thanks for sharing it @cryptotavares !)
* This talk: https://www.youtube.com/watch?v=Af4M8GMoxi4
* Probably more places that I am forgetting

Mrtenz (commented):
Overall this looks great. Maybe you could add a section on when to and when not to use snapshot tests, cause there may be some confusion around that.

Gudahtt:
I would love that! I felt that maybe this wasn't the place for it though? I will consider where that guidance would fit better.

docs/testing/overview.md
```diff
@@ -1,6 +1,52 @@
...
+
+What we consider "fast" depends on what the test does; what's fast for an end-to-end test might be slow for a unit test. But faster is always better, for all types of tests.
+
+### 7. Tests should have a high return on investment
+
+If a test doesn't align with the previous 6 principles, it's probably doing more harm than good. Consider deleting it.
+
+That doesn't mean we should skip writing tests with it's too difficult. On the contrary; we have high expectations for quality, we don't have the luxury of writing only the easiest most valuable tests. We need to aim for total coverage.
+
+But writing and maintaining bad tests is not the shortest path to complete test coverage. Our time is better spent writing good tests, and making investments that allow us to write more good tests. All code can be designed to be easy to test, and even the most complex test setup/teardown can be reasonably fast and easy to read/write with the right tools.
```
pnarayanaswamy, line 50: Would also add that Tests should be isolated:
A good test should not depend on the results or state of another test. Each test should clean up after itself and set up its environment, allowing for parallel or random execution.

docs/testing/overview.md
```diff
@@ -1,6 +1,52 @@
...
+
+## Testing principles
+
+What makes a good test? Here are some general principles.
+
+### 1. Tests should describe expectations
+
+What's the difference between a bug and a feature? A feature is expected, a bug is not.
+
+If it's not clear what the code-under-test is supposed to do, it won't be clear what to do when the test fails either. Clear expectations are essential for an effective test.
```
pnarayanaswamy, line 15: A test should ideally focus on one specific behavior or feature, reducing ambiguity in what failed if it does.

docs/testing/overview.md
Gudahtt: Great point! I would consider this as covered by "Tests should not break often". The reason I would say that we want tests to be isolated is that non-isolated tests break too often.

Gudahtt: There is also a tradeoff here to consider I think. For end-to-end tests, it can sometimes be worthwhile to have non-isolated tests for the sake of speeding up test execution time.

We had tests like this in the past and they were miserable to deal with, but with better tooling and better discipline about the expectations before/after each test, I think it can be a good strategy for minimizing test execution time.

docs/testing/overview.md
Gudahtt: Generally I think this is great advice, but I don't think it's precisely true to say that we want this for _all_ types of automated tests. End-to-end tests for example would typically have a broader scope, not targeted to a specific feature.

docs/testing/overview.md
```diff
@@ -1,6 +1,52 @@
...
+
+### 2. Tests should catch bugs
+
+This is the main purpose of tests. Take care to ensure that each test actually fails when the described expectation is not met.
+
+### 3. Tests should not break often
+
+It's expected that tests will break when we change the behavior or APIs under test. But if a test breaks often from refactors, or from changes to behavior or APIs not directly under test, it's sign that the test may be poorly designed.
```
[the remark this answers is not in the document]
Gudahtt: True... I referenced this elsewhere, but you're right, it sounds like I'm always blaming the test here. I will revise this.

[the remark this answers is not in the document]
Gudahtt: Clarified here: e293795

docs/testing/overview.md
```diff
@@ -1,6 +1,52 @@
...
+
+### 4. Tests should be easy to read
+
+A test isn't effective if we can't understand why it failed. The more readable a test is, the faster we can fix it.
+
+### 5. Tests should be easy to write
+
+Some behaviors are inherently more complex than others, increasing the difficulty of testing them. But in general we want tests to be easy to write, so that we can keep costs low.
+
+If a test is hard to write, consider that as a warning that you might be approaching the problem in a non-optimal way. Resist the urge to power through the work, and instead consider how redesigning the code-under-test or using a different testing strategy might save time.
```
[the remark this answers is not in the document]
Gudahtt: In these cases, I'd recommend protecting against major regressions using end-to-end tests and/or integration tests. If the test is sufficiently distant from the code, it will be easier to refactor/redesign. This can give you some safety while you work to make the code easier to test comprehensively.

I considered adding guidance like this here, but it felt a bit out of place. I was trying to keep these principles minimal, with specific guidance elsewhere. I am not sure where this advice would best fit though 🤔

docs/testing/overview.md
Gudahtt: This is covered already here as well: https://github.com/MetaMask/contributor-docs/blob/main/docs/testing/unit-testing.md#keep-tests-isolated

Gudahtt: About the tradeoffs for isolated end-to-end tests, this is described here: https://github.com/MetaMask/contributor-docs/blob/main/docs/testing/e2e-testing.md#guidelines-1

I don't completely agree with how it's discussed in that section, but I do agree with the basic point that there are tradeoffs to consider.

docs/testing/overview.md
```diff
@@ -1,6 +1,52 @@
...
+
+### 6. Tests should be fast
+
+Making tests faster has a dramatic improvement on productivity, reducing feedback loops.
+
+What we consider "fast" depends on what the test does; what's fast for an end-to-end test might be slow for a unit test. But faster is always better, for all types of tests.
```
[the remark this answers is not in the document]
Gudahtt: Good point. This is covered already in our extension e2e testing guidelines: https://github.com/MetaMask/contributor-docs/blob/main/docs/testing/e2e/extension-e2e-guidelines.md#wait-for-commands
But not in the mobile or generic e2e test guidelines. It probably should be moved up a level, as it applies to both clients.

docs/testing/overview.md
```diff
@@ -1,6 +1,52 @@
...
+
+This is the main purpose of tests. Take care to ensure that each test actually fails when the described expectation is not met.
+
+### 3. Tests should not break often
+
+It's expected that tests will break when we change the behavior or APIs under test. But if a test breaks often from refactors, or from changes to behavior or APIs not directly under test, it's sign that the test may be poorly designed.
+
+Test breakages are expensive. We should do what we can to minimize them.
+
+### 4. Tests should be easy to read
```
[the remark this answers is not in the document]
Gudahtt: Good advice! The first part about understanding the feature under test is hopefully covered by the first principle. The point about linking GitHub issues is a great practice, but feels too in-the-weeds for this section of the document.

docs/testing/overview.md
```diff
@@ -1,6 +1,52 @@
...
+
+### 1. Tests should describe expectations
+
+What's the difference between a bug and a feature? A feature is expected, a bug is not.
+
+If it's not clear what the code-under-test is supposed to do, it won't be clear what to do when the test fails either. Clear expectations are essential for an effective test.
+
+### 2. Tests should catch bugs
+
+This is the main purpose of tests. Take care to ensure that each test actually fails when the described expectation is not met.
```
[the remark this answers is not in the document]
Gudahtt: Great policy, though this feels like it would belong in a process document somewhere. Not sure exactly where it fits 🤔

docs/testing/overview.md
ccharly: IIRC, the "test isolation" was already documented here: https://github.com/MetaMask/contributor-docs/blob/main/docs/testing/unit-testing.md#keep-tests-isolated

IDK if we want to make reference to those more "specific" sections of the docs, but aiming to have systemic patterns for our tests and using the same "practices" is definitely valuable too. So if this overview is the "entrypoint" for writing good tests, it could be great to mention them somewhere? (I have no proposal for this though 🙈)

ccharly: ah, I didn't refresh my page before you posted your 2 last comments @Gudahtt sorry 🙈

docs/testing/overview.md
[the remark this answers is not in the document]
Gudahtt: Maybe I can include this in the last principle, where it talks about deleting tests / not writing bad tests. I can have a sentence at the end about this.

docs/testing/overview.md
cryptotavares: I actually like this being inlined... A testing overview doc should be the entry/starting point for devs when they are writing tests for the first time in our codebase. So it is important to highlight what are our transversal testing principles before diving into the different types of tests.

docs/testing/overview.md
cryptotavares: Agree with @Gudahtt here... Tests should reveal intent (and as so clearly describe the expectations). Then depending on the type of test, the intent can be beyond a specific feature or behaviour. Usually for unit tests, I would agree with the statement, but for e2e tests for example I would rather focus the tests on use cases (and those can expand through multiple features/behaviours).

docs/testing/overview.md
```diff
@@ -1,6 +1,52 @@
...
+
+What we consider "fast" depends on what the test does; what's fast for an end-to-end test might be slow for a unit test. But faster is always better, for all types of tests.
+
+### 7. Tests should have a high return on investment
+
+If a test doesn't align with the previous 6 principles, it's probably doing more harm than good. Consider deleting it.
+
+That doesn't mean we should skip writing tests with it's too difficult. On the contrary; we have high expectations for quality, we don't have the luxury of writing only the easiest most valuable tests. We need to aim for total coverage.
+
+But writing and maintaining bad tests is not the shortest path to complete test coverage. Our time is better spent writing good tests, and making investments that allow us to write more good tests. All code can be designed to be easy to test, and even the most complex test setup/teardown can be reasonably fast and easy to read/write with the right tools.
```
cryptotavares, line 49: This is such an important statement 👏

cryptotavares (dismissed):
Really love the outlined test principles.. they are high level enough, allowing us to follow them regardless of the type of test that we are implementing, but at the same time if the principles are followed, it is highly probable that we end up writing a good test (again regardless if we are writing a unit/integration/e2e test).

🚀 🚀 🚀

docs/testing/overview.md
seaona, line 1: great addition @Gudahtt :raised_hands: 
Overall LGTM. I was thinking on another principle: **Tests should be deterministic**
That's somehow implicit from the existing sections already, but do you think it might deserve its own point?

seaona (approved):

docs/testing/overview.md
Gudahtt: Deterministic is a good suggestion! But I hesitate to add it here for two reasons:

* The "tests should rarely break" principle already mostly captures this. The main reason a non-deterministic test is harmful is that it causes test breakages.
* Some tests might intentionally be non-deterministic!

  Ideally our code is deterministic, but if we were to have some non-deterministic behaviour, the test might also be non-deterministic. We of course make every effort to reduce unpredictability in our tests, but it's common practice to have some tests that are "more realistic but less reliable" that run nightly or just on the main branch, which makes a different tradeoff between "catching bugs" and "reducing test breakages" (i.e. they break more often but catch some bugs not caught elsewhere).

  This strategy should be approached with caution I think, as it is a potential time-sink. But it can be a reasonable tradeoff to make. This is why the principle "Tests should rarely break" was worded as intentionally "fuzzy" rather than being stated as an absolute "tests should never break in the absence of a behavior/API change or a bug".

Gudahtt: >"Do not test implementation details" -> users do not care about it, and neither should the tests.

This is a good suggestion 🤔 I have been thinking about re-wording the introduction where the goal is stated ("complete test coverage" is ambiguous and would mean different things to different people, and isn't the best representation of our goal anyway).

Taken more generally, this might be an inverse of the "tests should catch bugs" point. e.g. "Do not write tests that do not catch bugs". I will consider adding this clarification there.

Gudahtt:
@Mrtenz I've added an issue here to track the idea for adding more snapshot testing guidelines: https://github.com/MetaMask/contributor-docs/issues/115

docs/testing/overview.md
Gudahtt, line 15:
```suggestion
Each test should be verifying specific expectations. If it's not clear what the code-under-test is supposed to do, it won't be clear what to do when the test fails either. Clear expectations are essential for an effective test.
```

docs/testing/overview.md
Gudahtt, line 49:
```suggestion
However, writing and maintaining bad tests is not the shortest path to complete test coverage. Our time is better spent writing good tests, and making investments that allow us to write more good tests. All code can be designed to be easy to test, and even the most complex test setup/teardown can be reasonably fast and easy to read/write with the right tools.
```

seaona (approved):

docs/testing/overview.md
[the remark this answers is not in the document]
Gudahtt: I wasn't able to find a place for this. I'll look elsewhere for where this advice can live, in a later PR.

docs/testing/overview.md
Gudahtt: I wasn't able to find a nice way of fitting in a re-wording of the introduction, or the advice about not testing implementation details.

When I tried fitting in "Don't test implementation details" part, it felt a little too prescriptive to be in this section because it could be read as disallowing mockist/London-school/inside-out testing strategies (which certainly _can_ be effective, though they're certainly not my preference). From a certain point of view, what might be called "implementation details" by one person could be "the contract this code is meant to fulfill / the best representation of this code's responsibility" by another.

I do plan on updating the later parts of this document with a more explicit recommendation to prefer less mocking (e.g. a recommendation _against_ mockist/inside-out testing), because I don't think it's practical to apply that strategy to our legacy codebase. But for teams that want to use that strategy for their own isolated modules, it feels a bit heavy-handed to disallow it completely, or to recommend against it as a general principle.

PR #126, opened 2025-02-21
feat(4243): add doc for create an analystics event from extension

DDDDDanica:
Create a new contributor-docs explaining minimum knowledge required to create an analytics event in the MetaMask extension codebase.


gauthierpetetin:
Thanks a lot @DDDDDanica for kicking off the tutorial on analytics in Extension!
@NicolasMassart  since you have worked a lot on analytics, it would be great to have your insights as well, so I added you as a reviewer. We will also need to include content for analytics in Mobile, probably in the same PR. We can discuss this during the weekly review.

docs/create-analystics-event.md
```diff
@@ -0,0 +1,143 @@
+# Create Analytics Event
+
+## Overview
+
+This document outlines the process for creating a new analytics event in MetaMask (Extension and Mobile)
+
+## Prerequisites
+
+- Before you begin, please refer to this [Working with Event Metrics](https://www.notion.so/consensys/Working-with-Event-Metrics-3b0b5a4308e64649a54a2864886c0412) and make sure you understand the context of the analytics events.
```
NicolasMassart, line 9: Q: is this Notion page public for external contributors?

docs/create-analystics-event.md
```diff
@@ -0,0 +1,143 @@
...
+            property_two: 'value_two',
+          },
+        },
+      ],
+    })
+    .thenCallback(() => ({ statusCode: 200 }));
+}
+
+describe('YourNewEventName Analytics', () => {
+  it('should track the event with correct properties', async () => {
```
NicolasMassart, line 94:
```suggestion
  it('tracks event with properties', async () => {
```

docs/create-analystics-event.md
```diff
@@ -0,0 +1,143 @@
...
+        const events = await getEventPayloads(driver, mockedEndpoints);
+        assert.equal(events[0].properties.property_one, 'value_one');
+        assert.equal(events[0].properties.property_two, 'value_two');
+      },
+    );
+  });
+});
+```
+
+#### 2.4. Validate the Event
```
NicolasMassart, line 117: S: optional if user has access, but helps sometimes to detect issues, we should also check for the event in a Mixpanel dev dashboard.

docs/create-analystics-event.md
```diff
@@ -0,0 +1,143 @@
...
+   - Property values are being captured as intended
+
+![Analytics Event Request Process](./images/create-analystics-event-request.png)
+
+#### 2.5. Opt-out Handling
+
+The tracking system automatically respects user privacy settings:
+
+- Events are only tracked if the user has opted into MetaMetrics
+- Sensitive data is automatically anonymized
```
NicolasMassart, line 138: S: explain how this is anonymized. For instance on mobile we have:
- normal event: one event with all props and user id
- anonymous event: two events:
  - one normal event with only non sensitive props and user id (you know who but not what)
  - one anonymous event with sensitive props and an anonymous id (you know what but not who)

docs/create-analystics-event.md
```diff
@@ -0,0 +1,143 @@
...
+- You don't need to check opt-in status before calling trackEvent
+
+### 3. Implement the Event in the Mobile App
+
+TODO: Add instructions for the mobile app
```
NicolasMassart, line 143: S: probably better in another PR as it's very different given that we use the Segment SDK.
I would even create another page.
So we would have one general with schema and prerequisites, and one for mobile and one for ext.

docs/create-analystics-event.md
DDDDDanica: it is not public but critical for the context, wdyt? should we move this out

docs/create-analystics-event.md
DDDDDanica: Addressed dfaa2250e89183c257b75490dd7bab83968a61c0

docs/create-analystics-event.md
DDDDDanica: Gotcha, removed b4077531db905171ec59b1efb66607f82e0e7eaa, and actually schema is defined in https://github.com/Consensys/segment-schema/blob/main/CONTRIBUTING.md

docs/create-analystics-event.md
DDDDDanica: Good idea, added in d5727a2b7bae39d5613c511d4f07207c26e37c0d

docs/create-analystics-event.md
NicolasMassart: Can we copy it as MD in this repo?

docs/create-analystics-event.md
NicolasMassart: They have to access it (only for internal) using their Okta dashboard.

docs/create-analystics-event.md
NicolasMassart: Yes, another page, another PR

NicolasMassart (approved):
This looks good @DDDDDanica!
I approve even with the few possible improvements.
Feel free to fix them but maybe we are better merging this first and then improve it.

docs/create-analystics-event.md
DDDDDanica: There are a lot of charts stuff. IMO external contributor won't add a metric event, we can remove this. wdyt?

docs/create-analystics-event.md
DDDDDanica: Oh my bad, i mean link of Mixpanel document ...

kathleenmotley49-maker (approved):

PR #134, opened 2025-03-07
Update unit test naming guidelines

NicolasMassart:
# Test Naming Guidelines Update

Updates our test naming conventions to improve clarity and maintainability of our test suite.

## Key Changes

- Established clear guidelines for naming unit tests
- Added examples of good and bad practices
- Introduced principles for more descriptive and concise test names

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes to contributor guidelines; no runtime, security, or application code is affected.
> 
> **Overview**
> **Updates `docs/testing/unit-testing.md`** with clearer, more prescriptive guidance on how to name and structure unit tests, plus editorial tightening across the rest of the page.
> 
> The **`it` description** section is the main addition: eight numbered rules (present-tense verbs, outcome-focused wording, no filler like "should", avoid implementation details, etc.) and many new 🚫/✅ example pairs (tokens, redirects, errors, vague names). Snapshot tests now recommend names like **`matches rendered snapshot`** instead of **`should renders correctly`**, with complete example blocks.
> 
> Other sections are largely **reworded for brevity** (Jest intro, `describe` grouping, four-phase tests, isolation, `beforeEach` vs helpers, mocks, snapshots). Minor structural tweaks include bullet lists for private-code types, dropping the 💡 prefix on one heading, and removing one "Read more" link under test-as-specification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dfdcb070325f66bd771a5abdf4dd975255f3d0e5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/testing/unit-testing.md
+++ b/docs/testing/unit-testing.md
@@ -152,8 +152,6 @@ it('displays invalid address error', () => {
 });
 ```
 
----
-
 🚫 **Stating obvious successful outcomes**
 
 ```typescript
--- a/docs/testing/unit-testing.md
+++ b/docs/testing/unit-testing.md
@@ -1102,7 +1102,7 @@ describe('MyComponent', () => {
 
 ```ts
 describe('MyComponent', () => {
-  it('matches snapshot')
+  it('Matches rendered snapshot')
 ```
 
 Of course variants of this naming can be used to add some context, for instance:
--- a/docs/testing/unit-testing.md
+++ b/docs/testing/unit-testing.md
@@ -2,7 +2,7 @@
 
 ## Use Jest
 
-[Jest](https://jestjs.io/) is the testing framework of choice for unit tests at MetaMask. It offers a more comprehensive set of features out of the box over Mocha or Tape that make writing tests for production code very easy. Some of the useful features are:
+[Jest](https://jestjs.io/) is the testing framework for unit tests at MetaMask. Jest includes many built-in features that make it easier to write tests. Useful features include:
 
 - [Module mocks](https://jestjs.io/docs/next/mock-functions#mocking-modules)
 - [Timer mocks](https://jestjs.io/docs/next/timer-mocks)
@@ -12,7 +12,7 @@
 
 ## Colocate test files with implementation files
 
-Test files are much easier to find when they sit next to the code they test.
+Place test files next to the code they test. This makes test files easier to find.
 
 🚫
 
@@ -33,7 +33,10 @@ src/
 
 ## Use `describe` to group tests for the same function/method
 
-Wrapping tests for the same function or method in a `describe` makes it easier to spot them in a large test file and makes it possible to run them on their own (using `.only`). It also helps to establish the subject of each test (the "it") and therefore keeps it well described and focused.
+Wrap tests for the same function or method in a `describe` block. This provides three benefits:
+- Tests are easier to find in large test files
+- You can run only these tests using `.only`
+- The test subject (the "it") is clear and focused
 
 🚫
 
@@ -69,7 +72,7 @@ describe('KeyringController', () => {
 
 ## 💡 Use `describe` to group tests under scenarios
 
-If you have many tests that verify the behavior of a piece of code under a particular condition, you might find it helpful to wrap them with a `describe` block so you only need to specify that condition once. Use a phrase starting with `if ...` or `when ...` so as to form a sentence when combined with the test description.
+When multiple tests verify behavior under the same condition, wrap them in a `describe` block. This way, you specify the condition only once. Start the description with `if ...` or `when ...` to form a complete sentence with each test description.
 
 1️⃣
 
@@ -107,15 +110,12 @@ describe('when adding a token', () => {
 
 ## Use `it` to specify the desired behavior for the code under test
 
-As each test [has to focus on a single aspect of that behavior](#keep-tests-focused), its description must describe that behavior using these guidelines:
+As each test [has to focus on a single aspect of that behavior](#keep-tests-focused), its description must describe that behavior clearly:
 
-1. Start with an active verb in present tense
-2. Focus on what is being tested, not how
-3. Be explicit about the context when needed
-4. Keep names concise but descriptive
-5. Avoid "should", "when", and other unnecessary words
-6. Do not state obvious successful outcomes ("works successfully", "correctly")
-7. Do not list test parameters; describe what they represent instead
+1. Start with an active verb in present tense (e.g., "returns", "displays", "prevents")
+2. State the expected outcome or behavior directly
+3. Add contextual conditions only when necessary for clarity (e.g., "when session expires")
+4. Keep descriptions concise but complete
 
 ### Examples
 
@@ -241,14 +241,14 @@ it('accepts transaction up to maximum amount limit', () => {
 });
 ```
 
-The test description should communicate the expected behavior clearly and directly. Avoid:
+Based on the examples above, avoid these common pitfalls:
 
-- Repeating the name of the function or method being tested that is already in the "describe" section.
-- Using "should" at the beginning of the test name
-- Including implementation details (like function calls or internal methods)
-- Stating obvious successful outcomes ("works correctly", "successfully" or "gracefully")
-- Listing test parameters instead of what they represent
-- Using vague terms like "fail", "error", or "throw" without being specific about the exact behavior (error type)
+- **Filler words**: "should", "correctly", "successfully", "gracefully", "properly"
+- **Vague descriptions**: "test", "edge case", "works", "handles"
+- **Implementation details**: Describing function calls (e.g., "calls redirectTo") instead of behavior (e.g., "redirects to login")
+- **Generic error language**: "throws an error" instead of specific behavior (e.g., "throws InvalidPayloadError" or "prevents sending")
+- **Parameter lists**: Listing conditions instead of describing what they mean
+- **Redundancy**: Repeating the function name already in the `describe` block
 
 ### Read more
 
@@ -256,9 +256,9 @@ The test description should communicate the expected behavior clearly and direct
 
 ## Keep tests focused
 
-Tests are easier to reason about and maintain when they cover only one aspect of the intended behavior.
+Tests are easier to understand and maintain when they cover only one aspect of behavior.
 
-If you are using "and" in a test description, this could indicate the test is too large and may need to be broken up.
+If you use "and" in a test description, the test is probably too large. Split it into separate tests.
 
 🚫
 
@@ -282,15 +282,17 @@ it('returns the block number', () => {
 
 ## Don't directly test private code
 
-Private code is a part of an interface that is not intended to be used by consumers of that interface. This could mean:
+Private code is not intended for consumers of an interface. Private code includes:
 
-- a function or class not exported from a module
-- a method whose name starts with `#` (the syntax for [ECMAScript private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields))
-- a method whose name starts with `_` (an informal approach popular prior to private fields)
-- a method preceded by the `private` keyword in TypeScript
-- a function or method tagged with the `@private` JSDoc tag
+- Functions or classes not exported from a module
+- Methods that start with `#` ([ECMAScript private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields))
+- Methods that start with `_` (older convention before private fields)
+- Methods with the `private` keyword in TypeScript
+- Functions or methods tagged with `@private` in TSDoc
 
-Test code marked as private as though it were directly part of the public interface where it is executed. For instance, although you might write the following:
+**Do not test private code directly.** Instead, test the public methods that call the private code. Write tests as if the private code is part of the public method.
+
+For example, you might write this code:
 
 ```typescript
 // block-tracker.ts
@@ -324,7 +326,7 @@ export class BlockTracker extends EventEmitter {
 }
 ```
 
-this is what consumers see:
+Consumers (and your tests) can only see the public interface, which behaves as if the private methods were inlined:
 
 ```typescript
 // block-tracker.ts
@@ -350,7 +352,7 @@ export class BlockTracker extends EventEmitter {
 }
 ```
 
-and as a result, this is how the method ought to be tested:
+Therefore, test the public `stop` method by verifying all the behaviors that result from calling it, including the effects of the private methods:
 
 ```typescript
 describe('BlockTracker', () => {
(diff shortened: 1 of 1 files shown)


docs/testing/unit-testing.md
```diff
@@ -107,62 +107,67 @@ describe('when adding a token', () => {
...
 ```
 
+The test description should communicate the expected behavior clearly and directly. Avoid:
+
+- Repeating the name of the function or method being tested
+- Using "should" at the beginning of the test name
+- Including implementation details in the name
+- Stating obvious successful outcomes
+- Listing test parameters instead of what they represent
+- Using words like "fail", "error", or "throw" when the error is the expected behavior
```
Mrtenz, line 169: Not sure about this. Sometimes if we want to test validation behaviour for example, I'd write something like this:

```js
it('throws an error when ...', () => {
  // ...
});
```

mcmire: I would agree with this. I understand behavior can be described different ways, and sometimes maybe it's enough to be more abstract in the description. But I do find it useful, on the unit level, to include _some_ implementation details in the description, to be able to quickly understand the behavior. So this would be, what the unit under test a) returns, b) throws, or c) calls (and, if appropriate, in the context of a scenario such "...when passed A" or "...when B is set").

docs/testing/unit-testing.md
```diff
@@ -107,62 +107,67 @@ describe('when adding a token', () => {
...
+As each test [has to focus on a single aspect of that behavior](#keep-tests-focused), its description must describe that behavior using these guidelines:
 
-Do not repeat the name of the function or method in the name of the test.
-
-Do not use "should" at the beginning of the test name. The official Jest documentation [omits this word from their examples](https://jestjs.io/docs/next/getting-started), and it creates noise when reviewing the list of tests printed after a run.
+1. Start with an active verb in present tense
+2. Focus on what is being tested, not how
+3. Be explicit about the context when needed
+4. Keep names concise but descriptive
+5. Avoid "should", "when", and other unnecessary words
```
mcmire, line 116: Avoid "should" makes sense, but I do find it useful to highlight under which condition the behavior under test applies, if any, as it's easy to compare with the code being tested. Do you find this helpful or is this is not something that you tend to encounter?

docs/testing/unit-testing.md
```diff
@@ -107,62 +107,67 @@ describe('when adding a token', () => {
...
+6. Do not state obvious successful outcomes ("works successfully", "correctly")
+7. Do not list test parameters; describe what they represent instead
 
 ### Examples
 
-🚫
+🚫 Don't
```
mcmire, line 122: Something I've been doing for other guidelines is describe why something is good or bad. I didn't do it here but maybe we want to fix that? So, perhaps we could break up these examples into pairs. Something like:

~~~ markdown
**🚫 "should" is used in the test name**

```typescript
it('should successfully add token when address is valid and decimals are set and symbol exists', () => {
  // ...
});
```

**🚫 The test name describes the behavior as though it already exists**

```typescript
it('successfully adds token when address is valid and decimals are set and symbol exists', () => {
  // ...
});
```
~~~

...and then following with the next pair of examples, and the next, and so on.

docs/testing/unit-testing.md
```diff
@@ -107,62 +107,67 @@ describe('when adding a token', () => {
...
-```
 
-🚫
+it('works correctly when processing the transaction', () => {
+  // ...
+});
 
-```typescript
-describe('TokensController', () => {
-  it('addToken', () => {
```
mcmire, line 138: I wonder if we should still include this example or something like it. This is something I've seen in some really old tests, where people don't bother to write a description for the thing they're testing. Then again, maybe it's obvious, I don't know. Just thought I'd point it out.

NicolasMassart: yes definitely, will work on this

Rsed19901 (approved):
rsedr1990@gmail.com

docs/testing/unit-testing.md
NicolasMassart: Also I’d argue that test names should clarify what is being verified, not how it’s implemented. For example:

✅ redirects to login when session expires

❌ calls redirectTo('/login') when session expires

Implementation details like function calls or specific errors are better captured in the test body. This keeps test names stable even if internal logic changes, as long as the behavior remains consistent.

docs/testing/unit-testing.md
```diff
@@ -1097,7 +1102,7 @@ describe('MyComponent', () => {
 
 ```ts
 describe('MyComponent', () => {
-  it('render matches snapshot')
+  it('matches snapshot')
```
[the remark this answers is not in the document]
NicolasMassart: What about "Matches rendered snapshot"?
The expectation action first as a verb, then what should match, then what it should match.

docs/testing/unit-testing.md
```diff
@@ -107,71 +111,155 @@ describe('when adding a token', () => {
...
-As each test [should focus on a single aspect of that behavior](#keep-tests-focused), its description should describe that behavior. This description helps to anchor the purpose of the test, understand the intended behavior, and debug differences with the actual behavior that may occur down the road.
-
-Do not repeat the name of the function or method in the name of the test.
+As each test [has to focus on a single aspect of that behavior](#keep-tests-focused), its description must describe that behavior clearly:
 
-Do not use "should" at the beginning of the test name. The official Jest documentation [omits this word from their examples](https://jestjs.io/docs/next/getting-started), and it creates noise when reviewing the list of tests printed after a run.
+1. Start with an active verb in present tense (e.g., "returns", "displays", "prevents")
+2. State the expected outcome or behavior directly
+3. Add contextual conditions only when necessary for clarity (e.g., "when session expires")
+4. Keep descriptions concise but complete
```
mcmire: What do you think about moving the extra set of guidelines from after the list to examples to this section?

``` suggestion
Just as [each test must focus on a single aspect of the behavior](#keep-tests-focused), its description must also describe that behavior clearly.

To do this:

1. Start the description with an active verb in present tense (e.g., "returns", "displays", "prevents").
2. State the expected outcome or behavior directly.
3. Add contextual conditions only when necessary for clarity (e.g., "when session expires").
4. Keep descriptions concise but complete.
5. Avoid filler words ("should", "correctly", "successfully", "gracefully", "properly").
6. Avoid vague descriptions ("test", "edge case", "works", "handles").
7. Avoid implementation details ("calls redirectTo", "throws InvalidPayloadError").
8. Don't use the name of the function as the test description.

File: docs/testing/unit-testing.md [hunk 7]
```diff
@@ -195,15 +283,17 @@ it('returns the block number', () => {
...
 
-- a function or class not exported from a module
-- a method whose name starts with `#` (the syntax for [ECMAScript private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields))
-- a method whose name starts with `_` (an informal approach popular prior to private fields)
-- a method preceded by the `private` keyword in TypeScript
-- a function or method tagged with the `@private` JSDoc tag
+**Do not test private code directly.** Instead, test the public methods that call the private code. Write tests as if the private code is part of the public method.
 
-Test code marked as private as though it were directly part of the public interface where it is executed. For instance, although you might write the following:
+For example, you might write this code:
```
mcmire:
``` suggestion
Take this code for example:
```

docs/testing/unit-testing.md
```diff
@@ -815,14 +905,14 @@ function buildTokenDetectionController({
...
 
-- It makes tests harder to read. Tests that run under different scenarios may require different ways of being set up, but using a `beforeEach` hook unnecessarily assigns the same importance to all setup steps listed there. This forces the reader to peruse all of the setup code in order to distinguish the signal from the noise.
-- It makes writing tests for new scenarios difficult. The setup steps specified in the `beforeEach` section may not perfectly apply to future tests that cover different scenarios. In these cases, the author may be forced to take on a complex refactor to remove the `beforeEach` steps that don't apply or pursue workarounds which decrease consistency and readability.
+- **Makes tests harder to read.** Different tests may need different setup, but `beforeEach` assigns equal importance to all setup steps. Readers must read all the setup code to find what matters for each test.
+- **Makes writing new tests difficult.** The `beforeEach` setup may not fit new test scenarios. You may need complex refactoring to remove steps that don't apply, or use workarounds that hurt consistency and readability.
```
mcmire: I like the use of "it" before these bullet points, otherwise it sounds too clinical:

``` suggestion
- **It makes tests are harder to read.** Different tests may need different setup, but `beforeEach` assigns equal importance to all setup steps. Readers must read all the setup code to find what matters for each test.
- **It makes writing new tests more difficult.** The `beforeEach` setup may not fit new test scenarios. You may need complex refactoring to remove steps that don't apply, or use workarounds that hurt consistency and readability.

File: docs/testing/unit-testing.md
Comment (mcmire) [inline docs/testing/unit-testing.md]:
``` suggestion
Here's how we could improve this:

✅
```

mcmire (commented):
Sorry this has been open for so long, I just realized I already reviewed this a while back but I forgot about this 🤦🏻 

Thank you for this! As the original author of this guide, I fully admit I have a habit of using too many words. So I appreciate these changes. I had some comments/suggestions below, but they are all minor.

docs/testing/unit-testing.md
```diff
@@ -289,16 +379,16 @@ describe('BlockTracker', () => {
...
-2. **Exercise:** Executing the code under test
-3. **Verify:** Confirming that the code under test behaves in an expected manner
-4. **Teardown:** Returning the environment to a clean slate
+1. **Setup:** Configure the environment to run the code under test
+2. **Exercise:** Execute the code under test
+3. **Verify:** Confirm that the code behaves as expected
+4. **Teardown:** Return the environment to a clean state
 
-Be aware of the way that a test moves through these phases. It can assist readers in following the flow by making judicious use of line breaks to separate phases visually.
+Use empty lines to separate these phases visually. This helps readers understand the test flow.
```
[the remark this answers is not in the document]
NicolasMassart: Yes it's a suggestion but we can still ask for the simplest form of separation imo, be straightforward, and then if people really think it has to be different, they can as soon as it's clearer.
I hesitate to add "consider" as a guideline is already something you are not forced to use. "Consider" makes it feel even more optional...

docs/testing/unit-testing.md
```diff
@@ -576,7 +666,7 @@ describe('interpretMethodData', () => {
 });
 ```
 
-Even better, however, is to make use of [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection) to eliminate the need for a global. This allows you to create a fake version of the value in question within your test:
+A better approach is to use [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection) to remove the need for a global variable. This lets you create a fake value within your test:
```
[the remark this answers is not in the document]
NicolasMassart: Again, this is for ease of reading from non native readers

docs/testing/unit-testing.md
```diff
@@ -672,7 +762,7 @@ describe("interpretMethodData", () => {
 });
 ```
 
-[Instead of using hooks](#avoid-before-each-and-after-each), however, use a factory function which defines defaults and allows them to be overridden where they are needed:
+[Instead of using hooks](#avoid-before-each-and-after-each), use a factory function. This function defines default values and lets you override them when needed:
```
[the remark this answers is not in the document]
NicolasMassart: Same, for non native readers.

docs/testing/unit-testing.md
```diff
@@ -815,14 +905,14 @@ function buildTokenDetectionController({
...
+- **Makes tests harder to read.** Different tests may need different setup, but `beforeEach` assigns equal importance to all setup steps. Readers must read all the setup code to find what matters for each test.
+- **Makes writing new tests difficult.** The `beforeEach` setup may not fit new test scenarios. You may need complex refactoring to remove steps that don't apply, or use workarounds that hurt consistency and readability.
 
-Setup helper functions serve the same purpose as hooks without causing these problems. In this way, tests that need setup data for specific scenarios can specify them easily by passing options to the function, which communicates their importance to readers over other kinds of setup data in the context of the test. Plus, managing setup code becomes easier, as it can be refactored more easily if necessary.
+Setup helper functions solve these problems. Tests can pass options to the function to specify the setup they need. This shows readers what is important for each test. Setup code is also easier to refactor when needed.
```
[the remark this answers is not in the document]
NicolasMassart: Why not "solve"?
The guideline is not expected to be a novel with literary qualities or legal text.
It should be short, concise, precise yet easy to understand for anyone whatever their native language and english skills. Also it makes it easier for automated translator and AI.

docs/testing/unit-testing.md
```diff
@@ -542,7 +632,7 @@ describe('interpretMethodData', () => {
 });
 ```
 
-[Instead of using hooks](#avoid-before-each-and-after-each), one idea is to create a function wrapping your test which captures the current value of the global beforehand and restores the global to this value afterward:
+[Instead of using hooks](#avoid-before-each-and-after-each), create a function that wraps your test. This function captures the current global value before the test and restores it afterward:
```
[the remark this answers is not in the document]
mcmire: Actually I'm not sure I like this suggestion. We don't need to suggest a path that we don't want the reader to take.

docs/testing/unit-testing.md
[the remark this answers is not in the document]
mcmire: I added "still" to accompany the suggestion I made right before this one. But I decided against that, so this suggestion isn't needed.

docs/testing/unit-testing.md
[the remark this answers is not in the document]
mcmire: Anyway, I was mainly looking for these sentences to sound a little more natural. What about this:

``` suggestion
Setup helper functions solve these problems in two ways:

- Tests can pass options to the function to specify the setup they need, showing readers what is important for each test.
- Setup code is easier to refactor when needed.
```

docs/testing/unit-testing.md
mcmire: Thank you for checking me :)

docs/testing/unit-testing.md
```diff
@@ -263,7 +353,7 @@ export class BlockTracker extends EventEmitter {
 }
 ```
 
-and as a result, this is how the method ought to be tested:
+Therefore, test the public `stop` method by verifying all the behaviors that result from calling it, including the effects of the private methods:
```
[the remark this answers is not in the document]
mcmire: Another attempt:

``` suggestion
A test suite for this class might look like the following:
```

mcmire (commented):
Thanks for the discussion @NicolasMassart. Again I appreciate your attempts to make this more understandable and thanks for your patience. I think we are close on this.

docs/testing/unit-testing.md
```diff
@@ -195,15 +283,17 @@ it('returns the block number', () => {
...
+- Methods that start with `_` (older convention before private fields)
+- Methods with the `private` keyword in TypeScript
+- Functions or methods tagged with `@private` in TSDoc
 
-- a function or class not exported from a module
-- a method whose name starts with `#` (the syntax for [ECMAScript private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields))
-- a method whose name starts with `_` (an informal approach popular prior to private fields)
-- a method preceded by the `private` keyword in TypeScript
-- a function or method tagged with the `@private` JSDoc tag
+**Do not test private code directly.** Instead, test the public methods that call the private code. Write tests as if the private code is part of the public method.
```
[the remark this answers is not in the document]
mcmire: For more context, I made this suggestion because I felt that the introduction was a bit too jarring. But I just realized that this may not be simple enough:

``` suggestion
Write tests for the public part of an interface only. Do not test code that is not intended for consumers to use, such as:

- Functions or classes not exported from a module
- Methods that start with `#` ([ECMAScript private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields))
- Methods that start with `_` (older convention before private fields)
- Methods with the `private` keyword in TypeScript
- Functions or methods tagged with `@private` in TSDoc
```

lannathompson65-arch (approved):

lannathompson65-arch (approved):

mcmire (approved):
Thanks for the changes! LGTM.

PR #140, opened 2025-04-15
Add TypeScript guideline for minimizing selector input types

### LINKED ISSUE #139: Add guideline to TypeScript article regarding selector typing
- See https://github.com/MetaMask/metamask-extension/pull/29094

MajorLift:
- Closes https://github.com/MetaMask/contributor-docs/issues/139

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -1075,20 +1075,24 @@ A selector function that directly queries state properties should define its inp
 
 ```typescript
 const selectTodos = (state: RootState) => {
-  const { todoSlice: { todosA, todosB } } = state;
+  const {
+    todoSlice: { todosA, todosB },
+  } = state;
   return { todosA, todosB };
-}
+};
 ```
 
 ✅
 
 ```typescript
-const selectTodos = (state: { 
-  todoSlice: Pick<RootState['todoSlice'], 'todosA' | 'todosB'>
+const selectTodos = (state: {
+  todoSlice: Pick<RootState['todoSlice'], 'todosA' | 'todosB'>;
 }) => {
-  const { todoSlice: { todosA, todosB } } = state;
+  const {
+    todoSlice: { todosA, todosB },
+  } = state;
   return { todosA, todosB };
-}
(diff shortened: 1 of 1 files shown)


docs/typescript.md
```diff
@@ -1064,3 +1064,89 @@ async function removeAccount(address: Hex): Promise<KeyringControllerState> {
...
+
+Eventually, many selectors end up being defined with a state type that is close to or equal to the entire Redux state.
+
+Paradoxically, selectors that only need access to very few properties end up needing to have the widest state type, because they tend to be merged into the most selectors across several nested levels.
+
+- **Atomic:**
+
+When selectors are actually invoked, including in test files, it's not always practical to prepare and pass in a very large state object.
+
+It's both safer and more convenient to restrict the state being passed into the selector to the minimum size required for the selector to function.
```
davidmurdoch, line 1144: Do you mean restrict the state _type_ here, not the `state`?

MajorLift: Thanks that's a crucial distinction. I'll fix it as follows:

```suggestion
It's both safer and more convenient to restrict the type of the state object being passed into the selector to the minimum size required for the selector to function. Note that this does not prevent the selector from accepting a larger state object.
```

MajorLift: Thanks that's a crucial distinction. I'll fix it as follows:

```suggestion
It's both safer and more convenient to restrict the state argument type of the selector to the minimum size required for the selector to function. Note that this does not prevent the selector from accepting a larger state object, but it does allow the selector to function with a state object that is less than complete, as long as it satisfies the input argument type.
```

davidmurdoch: perfect!

GlaudineR (approved):

cryptodev-2s (approved):
LGTM!

sephrimehran11-crypto:
My wallet crypto address0x4bbeEB066eD09B7AEd07bF39EEe0460DFa261520
Water key pelees

PR #170, opened 2026-04-14
TypeScript guidelines: Document `any` requirement for callback parameters subject to bidirectional assignment

MajorLift:
[View rendered version](https://github.com/MetaMask/contributor-docs/blob/jongsun%2Ftypescript%2Fcallback-any-contravariance/docs%2Ftypescript.md#any-is-required-for-callback-parameter-types-constrained-by-bidirectional-assignment)

---

> [!NOTE]
> **Low Risk**
> Low risk documentation-only change that clarifies when `any` is acceptable for callback parameters under bidirectional function-type constraints; no runtime or type-checking behavior is modified.
> 
> **Overview**
> Adds a new TypeScript guideline section that **carves out a narrow exception** for using parameter `any` in callbacks that sit in a *bivariant* position between two **fixed, irresolvable** function-type constraints.
> 
> Includes examples distinguishing *redesignable* vs *fixed* constraints, guidance to annotate `eslint-disable` comments with the acceptance criteria, and an appendix-style derivation explaining why `unknown`/`never` fail under `--strictFunctionTypes`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c4909644da68e5b5f0beb37f312365d366c4622. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -883,7 +883,7 @@ export class ComposableController<
 
 #### `any` is required for callback parameter types constrained by bidirectional assignment
 
-If a callback needs to be a valid _assignee_ of a narrower callback type, while also being _assignable_ to a wider callback type, its parameters must be typed with `any`. 
+If a callback needs to be a valid _assignee_ of a narrower callback type, while also being _assignable_ to a wider callback type, its parameters must be typed with `any`.
 
 **Example <a id="example-f2a3b7d1-9e4c-4f8a-b6c2-1d8e5a3c9f7b"></a> ([🔗 permalink](#example-f2a3b7d1-9e4c-4f8a-b6c2-1d8e5a3c9f7b)):**
 
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -881,13 +881,38 @@ export class ComposableController<
   const controllerMessenger = ControllerMessenger<any, any>;
   ```
 
-#### `any` is required for callback parameter types constrained by bidirectional assignment
+#### `any` is acceptable for callback parameters caught between two irresolvable function-type constraints
 
-If a callback needs to be a valid _assignee_ of a narrower callback type, while also being _assignable_ to a wider callback type, its parameters must be typed with `any`.
+A callback's parameters may be typed as `any` if three conditions all hold:
+
+1. **Bivariant position:** The callback must be _assignable_ to a wider function type and an _assignee_ of a narrower function type.
+2. **Irresolvable constraints:** No concrete type satisfies both directions, because the "wider" function type is not a supertype of the "narrower" function type, which creates a paradox.
+3. **Fixed constraints:** Neither constraint can be redesigned without affecting downstream callers or introducing semantic or structural inaccuracies.
+
+🚫 When the constraints are redesignable, the contravariance error should not be suppressed with `any`, as it legitimately signals a broken design:
+
+**Example <a id="example-a3c5e7f1-8b2d-4a6c-9e0f-1b3d5e7a9c2b"></a> ([🔗 permalink](#example-a3c5e7f1-8b2d-4a6c-9e0f-1b3d5e7a9c2b)):**
+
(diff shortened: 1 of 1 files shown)


docs/typescript.md
```diff
@@ -881,6 +881,58 @@ export class ComposableController<
   const controllerMessenger = ControllerMessenger<any, any>;
   ```
 
+#### `any` is required for callback parameter types constrained by bidirectional assignment
+
+This shape arises whenever a callback value is constrained on both sides of an assignment by _different_ external function types. Real-world instances:
+
+- A `coerces` map sitting between a library signature and a caller's own config — see [`metamask-extension#41104 (r3045807022)`](https://github.com/MetaMask/metamask-extension/pull/41104#discussion_r3045807022).
+- A messenger's `registerActionHandler` slot typed `(...args: any[]) => any`: it receives strongly-typed handler callbacks inward at registration _and_ is invoked with strongly-typed argument tuples outward at dispatch. `unknown[]` fails registration; `never[]` fails dispatch.
```
MajorLift:
```suggestion
- A messenger's `registerActionHandler` slot typed `(...args: any[]) => any`: it receives strongly-typed handler callbacks inward at registration _and_ is invoked with strongly-typed argument tuples outward at dispatch. `unknown[]` fails registration; `never[]` fails dispatch.
- A `coerces` map sitting between a library signature and a caller's own config — see [`metamask-extension#41104 (r3045807022)`](https://github.com/MetaMask/metamask-extension/pull/41104#discussion_r3045807022).
```

docs/typescript.md
```diff
@@ -881,6 +881,60 @@ export class ComposableController<
...
+takesWide(f2); // ✗ 'string' not assignable to 'never'
+
+// ✅ `any` — satisfies both directions
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let f3: (x: any) => void;
+takesWide(f3); // ✓
+f3 = givesNarrow; // ✓
+```
+
+> **Note:** The `eslint-disable` is intentional. `any` here is _not_ infectious: it is scoped to a single callback's parameter position and does not propagate to callers. The enclosing external APIs re-impose their own parameter types at each use site, so type safety is preserved where values actually flow. Prefer `unknown` or `never` when only one direction applies; reach for `any` only when both apply to the same parameter position.
```
Gudahtt: I'm not sure about how this guideline is framed.

This note says that this is safe because "The enclosing external APIs re-impose their own parameter types at each use site", and indeed that is _why_ this pattern is safe in the rare case where we've used it.

But this statement is not true of all situations where you might be tempted to use `any` to get around incompatible type expectations. "callback parameter types constrained by bidirectional assignment" may describe scenarios where there is no "enclosing external API" with genuinely safe types, where this error is a legitimate type error that indicates a broken design.

Gudahtt: I'm not totally sure what to suggest though. I am having trouble describing the safe scenarios here.

It's more than just a matter of encapsulation. We could encapsulate an instance of an invariant type causing a type error, and it still be a legitimate error. We need to be confident the error would never occur (e.g. we'll never attempt to pass in a parameter that doesn't match the callback parameter type, and we'll never return a value that doesn't match the callback return type)

MajorLift: Yes I'll clarify that sometimes we can't "re-impose types" and `any` won't be safe to use. 

If it's possible to re-type either constraint to resolve an incompatible configuration without affecting downstream callers or introducing semantic inaccuracies, the type error does signal broken design and we shouldn't use `any` to suppress it.

Will push a new draft based on this.

lannathompson65-arch (commented):
Done

lannathompson65-arch (commented):
Amazing

lannathompson65-arch (approved):

PR #172, opened 2026-06-03
Document direct-value threshold feature flags

pedronfigueiredo:
## Description

Updates the remote feature flag documentation to explain direct-value threshold entries using `thresholdVersion: 2`.

## Changes

- Documents the legacy threshold output shape as `{ name, value }`.
- Adds a direct-value threshold example for selectors that read root-level object fields.
- Clarifies that non-threshold object flags and version-based object flags do not need `thresholdVersion`.
- Updates the composed version + threshold example to place `thresholdVersion: 2` inside each threshold entry.

Related to this core PR https://github.com/MetaMask/core/pull/8908

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime, auth, or data-handling impact.
> 
> **Overview**
> Updates **`docs/remote-feature-flags.md`** to document **direct-value threshold entries** (`thresholdVersion: 2`) and how they differ from the legacy threshold output.
> 
> The doc now states that entries **without** `thresholdVersion` resolve to the legacy **`{ name, value }`** wrapper, and that selectors should use **`.value`** in that case. A new section explains **`thresholdVersion: 2`**, which returns the selected entry’s **`value` at the root** (like a plain object flag), and recommends **`thresholdName`** instead of **`name`** for entry labels so config stays clear without colliding with fields inside the value object.
> 
> It also notes that **version-only object flags** do not use `thresholdVersion`, and refreshes the **version + threshold** composed example to use **`thresholdName`** / **`thresholdVersion: 2`** on each threshold entry, with a note that omitting v2 keeps the legacy wrapper shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6131eb606099bdefa7278a3249d787db49a48446. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/remote-feature-flags.md
+++ b/docs/remote-feature-flags.md
@@ -83,6 +83,60 @@ Choose the appropriate feature flag type based on your needs:
 ]
 ```
 
+Threshold entries without `thresholdVersion` use the legacy output shape. After the controller selects the matching threshold entry, the processed flag is:
+
+```json
+{
+  "name": "feature is ON",
+  "value": true
+}
+```
+
+Use this shape when selectors read the selected flag data from `.value`.
+
+#### Direct-value threshold entries
+
+Use `thresholdVersion: 2` when a threshold-based flag should expose the selected `value` directly, matching the shape of a non-threshold object flag. This is useful when converting an existing object flag to threshold rollout and existing selectors read root-level fields such as `flag.enabled`.
+
+```json
+[
+  {
+    "thresholdName": "feature is ON",
+    "thresholdVersion": 2,
+    "scope": {
+      "type": "threshold",
+      "value": 0.3
+    },
+    "value": {
+      "enabled": true,
+      "minimumVersion": "13.10.0"
+    }
+  },
+  {
+    "thresholdName": "feature is OFF",
+    "thresholdVersion": 2,
+    "scope": {
+      "type": "threshold",
+      "value": 1
+    },
+    "value": {
+      "enabled": false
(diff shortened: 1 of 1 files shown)


docs/remote-feature-flags.md
```diff
@@ -83,6 +83,60 @@ Choose the appropriate feature flag type based on your needs:
...
+Use this shape when selectors read the selected flag data from `.value`.
+
+#### Direct-value threshold entries
+
+Use `thresholdVersion: 2` when a threshold-based flag should expose the selected `value` directly, matching the shape of a non-threshold object flag. This is useful when converting an existing object flag to threshold rollout and existing selectors read root-level fields such as `flag.enabled`.
+
+```json
+[
+  {
+    "thresholdName": "feature is ON",
```
DDDDDanica, line 106: Can we add some explanation about `thresholdName` and what `thresholdName` is versus `name`, whether it's required when using `thresholdVersion: 2`, and how they differ?

Something like 

When using `thresholdVersion: 2`, use `thresholdName` instead of `name` to label the entry. Because the processed output is the raw value (no wrapper), the field is renamed to avoid collision with any name key in the value object itself.

pedronfigueiredo: I added a clarifying paragraph about `thresholdName`

DDDDDanica (approved):

